PR #1083, opened 2026-04-15
fix(event-store): merge sidecar events into query() reads (#1082)

### LINKED ISSUE #1082: bug: sidecar-mode events are invisible to materializers and queries — silently breaks event-sourced views for non-primary MCP server instances
## Summary

When `exarchos_workflow set` runs on an MCP server instance that is in sidecar mode (PID lock held by another instance), it writes the resulting `state.patched` and `workflow.transition` events to `{streamId}.hook-events.jsonl` as designed. But:

1. `exarchos_event query` does not read sidecar files.
2. CQRS materializers (delegation-readiness-view, code-quality-view, etc.) only consume events from the main store.
3. There is no automatic merge from sidecar → main.
4. `exarchos_workflow set` returns `success: true` with no degraded-mode signal (only `exarchos_event append` returns `sequencePending: true`).

**Result:** every event-sourced gate downstream of `set` (e.g. `prepare_delegation`) sees a stale view and reports the workflow as not-ready, even though `state.json` is correctly updated. Affects any user running multiple concurrent Claude Code sessions or multiple MCP server instances against the same state directory — a normal multi-session use case.

## Reproduction

1. Start two Claude Code sessions, both with the exarchos plugin enabled. Session A's MCP server acquires the PID lock; session B enters sidecar mode.
2. In session B, run a feature workflow: `ideate` → `plan` → `plan-review` with `exarchos_workflow set` calls including `phase` transitions and `planReview.approved: true`.
3. Run `exarchos_orchestrate prepare_delegation`. It reports:
   ```
   blockers: ["plan not approved", "no task.assigned events found", "Plan artifact is missing"]
   ```
4. Run `exarchos_workflow get` for the same `featureId` — it correctly shows `planReview.approved: true`, `artifacts.plan` populated, `phase: "delegate"`, full tasks array.
5. Inspect `~/.claude/workflow-state/{streamId}.hook-events.jsonl` — the events are there (`state.patched`, `workflow.transition`, `workflow.started`), just not visible to the materializer.

## Expected

Either:

- Materializers and `exarchos_event query` read from sidecar files (preferred — fully transparent), OR
- Primary process auto-merges sidecar files at a regular interval / on next write (similar to a SQLite WAL checkpoint), OR
- Sidecar-mode instances refuse `set` and other event-sourced operations with a clear error so callers know to fall back / retry on the primary.

## Actual

Sidecar events are stranded indefinitely. `set` returns success. Downstream gates fail mysteriously without any operator-visible signal that we're in degraded mode.

## Severity

**HIGH.** The exarchos plugin is designed for multi-session use, and this defect silently breaks the `delegate` → `review` → `synthesize` pipeline for any non-primary session. Discovered while dogfooding a real workflow (`ideate` → `plan` → `delegate`) where the user could not proceed past `prepare_delegation` despite three correct `set` calls — see the evidence section below.

## Suggested fix

1. **Short-term (read fix):** have `EventStore.query()` and the materializer's event-stream readers walk both `{stream}.events.jsonl` AND `{stream}.hook-events.jsonl`, sorted by timestamp.
2. **Medium-term (merge):** have the primary instance run a periodic merge that drains sidecars, assigns sequence numbers, and appends to the main store.
3. **Independent (UX):** have `exarchos_workflow set` propagate a `sidecarPending: true` ack field when the underlying append landed in sidecar, so callers can detect degraded mode.

## Related

- #987 (closed) — JSONL/SQLite dual-write gap for team lifecycle events. Same family of materializer/storage divergence.
- #1001 (closed) — `workflow set` auto-emit behavior clarification.

## Evidence (from a real session)

Workflow `cli-vs-mcp-facade-analysis`, three concurrent MCP server PIDs:

```
PID 11535  (started 18:21, holds .event-store.lock)
PID 12282  (started 18:21, sidecar mode)
PID 82138  (started 18:58, sidecar mode)
```

Sidecar file `~/.claude/workflow-state/cli-vs-mcp-facade-analysis.hook-events.jsonl` contained:

```
4  state.patched           ← from each `set` call with updates
3  workflow.transition     ← from ideate→plan, plan→plan-review, plan-review→delegate
1  workflow.started        ← from init
1  workflow.compound-entry
```

Main file `~/.claude/workflow-state/cli-vs-mcp-facade-analysis.events.jsonl` contained only 10 `gate.executed` events from orchestrate handlers (which use a different code path that bypasses sidecar).

`prepare_delegation` reported the workflow as not ready despite `exarchos_workflow get` returning fully correct state. The user could not proceed to delegation without one of:

- killing the lock-holder PID
- manual sidecar replay into the main JSONL
- restarting the session

None of these are acceptable as a normal user workflow.

rsalus:
## Summary

Fixes #1082. `EventStore.query()` was sidecar-blind: it read only `{streamId}.events.jsonl` while sidecar-mode writes land in `{streamId}.hook-events.jsonl`. The sidecar merger only runs once at primary startup (`src/index.ts:244-249`), so any second Claude Code session could call `exarchos_workflow set` successfully but `prepare_delegation` would report stale blockers because materializers never saw the `state.patched` / `workflow.transition` / `task.assigned` events. This PR ships two coupled fixes.

### Tier 1 — read-fix in `EventStore.query()`

- `query()` detects the sidecar file and, when present, collects main events (offset/limit stripped), reads sidecar lines, sorts by timestamp, assigns synthetic sequences continuing from the true main-stream max, and applies `type` / `since` / `until` / `sinceSequence` filters uniformly to the merged stream. Offset/limit applied last on the combined result.
- Fast path when no sidecar file exists is unchanged (bench-safe, early termination preserved).
- Sidecar events are normalized to the full `WorkflowEvent` shape (populates `streamId`, `schemaVersion`, synthetic `sequence`) so downstream consumers don't special-case them.
- Guards `initializeSequence` from running on a sidecar-only stream to avoid a spurious ENOENT warning.
- RCA: `docs/rca/2026-04-14-sidecar-event-visibility-1082.md`.

### Tier 3 — `sidecarPending` ack on workflow handlers

- `exarchos_workflow init`, `set`, and `checkpoint` now surface `sidecarPending: true` in their response `data` when the underlying event append landed in the sidecar file (the store is in sidecar mode). Mirrors the `sequencePending` precedent on `exarchos_event append` (`event-store/tools.ts:15`).
- Addresses a DIM-2 (Observability) HIGH finding: before Tier 3, these handlers returned `{success: true}` with no signal that writes were routed to the sidecar — textbook silent fallback. Tier 1 made sidecar events visible at read time; Tier 3 makes the write-side degradation visible at ack time.
- Renames `SidecarMode_QueryStillWorksFromJsonl` → `SidecarMode_QueryReturnsMainEventsWhenNoSidecarWrites`. The original name pinned the bug as correct behavior (DIM-5 misleading test name post-fix).

### Out of scope — follow-up issue (Tier 2)

- **Live primary-drain** of sidecar files so synthetic sequences eventually become real and sidecar backlogs don't accumulate during long primary uptime. Tier 1 restores correctness; Tier 2 eliminates the drift surface. Tracked separately with full axiom framing (DIM-7 HIGH unbounded buffer, DIM-1 MEDIUM split-brain, DIM-4 HIGH test fidelity cost).

## Test Plan

- [x] **Tier 1:** 6 new regression tests in `src/event-store/store.test.ts` covering merge, timestamp ordering across multiple sidecar writes, type filter, idempotency-key preservation, `sinceSequence` against synthetic sequences, sidecar-only stream.
- [x] **Tier 3:** 8 new tests in `src/workflow/tools.sidecar.test.ts` covering init/set/checkpoint × {sidecar, normal} × {phase, field-update, null-eventstore} scenarios.
- [x] RED confirmed for both tiers before implementation; GREEN on all new tests after.
- [x] Full MCP server test suite: **4992 passed / 21 skipped / 0 failed** (was 4984 pre-Tier 3 — matches exactly 8 new tests).
- [x] Root test suite: **540 passed / 4 skipped / 0 failed**.
- [x] Typecheck clean at root and in MCP server.
- [x] `npm run skills:guard` passes.
- [x] Manual multi-session reproduction from the issue body: primary + sidecar EventStore, sidecar appends `state.patched` + `task.assigned`, materializer-style consumer over `query()` output now sees `planReview.approved: true` and a non-zero task count.


rsalus:
## Shepherd iteration 1 — CodeRabbit review addressed

Pushed `5cebedc2`. Per-finding response below.

### Addressed

| Finding | Resolution |
|---|---|
| `store.ts:575-579` backend early-return bypasses sidecar merge (major) | `query()` no longer short-circuits when `this.backend` is set. The sidecar-presence check now runs first; if a sidecar file exists, backend `queryEvents()` output is merged with sidecar events through the same path that handles local JSONL. Backend-mode stores with a concurrent sibling in sidecar mode no longer lose visibility. |
| `store.ts:599-633` concat applies offset/limit in file order (major) | Replaced `[...main, ...sidecar]` with a proper two-way `mergeByTimestamp()`. Both inputs are individually time-ordered (main by sequence = insertion order, sidecar explicitly sorted in `readSidecarForQuery`), so the merge is linear. Offset/limit apply after the merge. Ties break toward main (sidecar timestamps are never earlier than concurrent main events). New regression test `SidecarMode_QueryInterleavesMainAndSidecarByTimestamp` locks in the interleave semantic. |
| `store.ts:608-627` stale `mainMax` cache | `query()` now calls `initializeSequence()` whenever the main file exists, re-reading the on-disk counter before computing synthetic sidecar sequences. A concurrent sibling append no longer causes the cache to lag. |
| `tools.sidecar.test.ts:74-179` no-op sidecar ack coverage | Added `handleSet_SidecarMode_NoOpUpdateOmitsSidecarPending` — empty `updates` emits no event, so `sidecarPending` must be absent even in sidecar mode. |
| RCA materializer risk claim | Rewrote the Tier 1 Risk paragraph. The incremental materializer at `servers/exarchos-mcp/src/views/materializer.ts:155` uses `sequence > highWaterMark` filtering, so synthetic sidecar sequences advance the HWM and can cause real sequences to be silently dropped after the primary drains. This is a live correctness surface (not a theoretical caveat) until Tier 2. Documented two mitigations. |

### Declined with rationale

| Finding | Rationale |
|---|---|
| `tools.sidecar.test.ts:33-35` use non-self PID for "sibling" simulation | `acquirePidLock` branches on `isPidAlive(existingPid)` — only a *live* PID triggers sidecar fallback; a fabricated "sibling" offset would frequently map to a non-existent PID and cause the lock to be reclaimed as stale, making the test flaky or a silent no-op. `process.pid` is the only PID guaranteed alive for the test's duration. Updated the comment to explain *why* we write our own PID rather than framing it as a sibling simulation. |
| RCA YAML frontmatter required | The repo's existing RCAs (`docs/rca/2026-04-10-*.md`, `docs/rca/2026-02-25-*.md`) all open with `# RCA: …` — there is no frontmatter convention in this directory. Happy to introduce one if that's desired, but doing it in this PR would be a drive-by unrelated to #1082. |

### Verification
- MCP server suite: **4993 passed / 21 skipped / 0 failed** (+1 for the no-op ack test, timestamp-interleave regression).
- Root suite: **540 passed / 4 skipped / 0 failed**.
- Typecheck clean at root and in `servers/exarchos-mcp/`.

rsalus:
## Shepherd iteration 2 — follow-up review addressed

Pushed `b639d27d`. Per-finding response below.

### Addressed

| Finding | Resolution |
|---|---|
| CodeRabbit @ `store.ts:658` — sort `mainEvents` before merge (major) | `mergeByTimestamp` assumed both inputs were time-ordered, but JSONL/backend preserve stream order (not timestamp order) and `append()` accepts caller-supplied timestamps (lines ~330, ~364), so a backfilled main event can sit out of timestamp order. `query()` now sorts `mainEvents` by timestamp with a stable sequence tie-break before merging, so offset/limit slice the correct window. |
| CodeRabbit @ `store.ts:655` — backend `mainMax` not JSONL truth (major) | Replaced the `initializeSequence` call (which short-circuits to `backend.getSequence()` in backend mode) with a new `readJsonlMaxSequence()`. It prefers `.seq` (O(1)) cross-validated against JSONL line count, and only falls back to the backend when no local JSONL exists. Because backend dual-write is explicitly best-effort (`persistAndReplicate`), a lagging backend would otherwise let synthetic sidecar sequences collide with already-durable main events. |

### Scoped to Tier 2 (out of scope here)

| Finding | Rationale |
|---|---|
| Sentry @ `store.ts:804` — synthetic sidecar sequences are non-deterministic across queries (HIGH) | Real concern but a semantic change. Today sidecar synthetic sequences are `mainMax + i`; if the primary appends between two queries, the same sidecar event shifts (`N → N+1`), and a poller using `sinceSequence` can see duplicates. A stable scheme (fixed large offset, or persist-on-first-query) would break the contract asserted by the existing tests (`events[i].sequence` contiguous with main). The correct long-term fix is Tier 2 (live primary drain) — once sidecar events acquire real sequences on drain, the synthetic-sequence surface goes away entirely. In the meantime, consumers in sidecar-present environments should dedupe by `idempotencyKey` (preserved on synthetic events, see `readSidecarForQuery`) rather than relying on `sinceSequence` identity. Will add an explicit caveat to the RCA and issue #1082 tracker. |

### Verification
- MCP server suite: **4994 passed / 21 skipped / 0 failed** (+1 interleave regression from iter 1).
- Root suite: **540 passed / 4 skipped**.
- Typecheck clean.

Also previewing the Tier 2 issue scope — the Sentry finding is exactly the risk we named up front ("synthetic sequences until drain makes them real") and I'd rather keep the fix coherent with that scope than patch piecemeal.

rsalus:
## Shepherd status — ready for human review

Pushed `c4ffebf7` (one-line defensive comment on the sidecar type cast — CodeRabbit's only new feedback on iter 2 was this optional nitpick).

### Resolution summary

| Iteration | Commits | Findings addressed | Findings declined |
|---|---|---|---|
| 1 (`5cebedc2`) | merge-path + timestamp merge + interleave test + mainMax refresh + no-op ack test + RCA materializer-risk correction | 5 | 2 (test-PID framing, RCA frontmatter) |
| 2 (`b639d27d`, `a9a5c240`) | sort `mainEvents` defensively; `readJsonlMaxSequence` as JSONL-truth source; RCA caveat for synthetic-seq non-determinism | 2 | 1 Sentry HIGH (scoped to Tier 2) |
| 3 (`c4ffebf7`) | type-cast trust-boundary comment | 1 (CodeRabbit nitpick, optional) | — |

### Current state
- **CI: all green** — benchmark, Detect Changes, MCP Server, CI Gate, Seer Code Review.
- **CodeRabbit iter 2 review: `COMMENTED`** (not CHANGES_REQUESTED) — the sticky `CHANGES_REQUESTED` review-decision carries over from the first review until a maintainer dismisses or re-approves.
- **No open actionable findings** that this PR's scope (Tier 1 read-fix + Tier 3 ack) could reasonably address. The Sentry HIGH on synthetic-sequence stability is exactly the problem Tier 2 (live primary drain) was carved out to fix; documented as a caveat in the RCA and as consumer guidance (use `idempotencyKey` dedup in sidecar-present environments).

Ready for maintainer review / approval. After merge, I'll pick up the Tier 2 follow-up in a separate workflow.

rsalus:
## Shepherd iteration 3 — Sentry MEDIUM addressed

Pushed `ac49a25e`.

**Sentry @ `tools.ts:843-844` — `sidecarPending` false positive on v1 workflow field updates (MEDIUM):**
Valid. The `emittedEvents` heuristic checked `updateKeys.length > 0` without verifying `isEventSourced(state)`, so v1 workflows received `sidecarPending: true` even though no `state.patched` event was emitted (the event-first path is v2-only). Fixed by gating `patchEmitted` on the same `isEventSourced && eventStore != null` predicate the event-append path uses (line 699-702). Added regression test `handleSet_SidecarMode_V1WorkflowFieldUpdateOmitsSidecarPending`.

MCP suite: **4995 passed / 21 skipped / 0 failed** (+1 v1 regression test).

rsalus:
@coderabbitai Approve

PR #1112, opened 2026-04-16
fix(mcp): Windows CLI no-op — normalize path separators in isDirectExecution (#1085)

### LINKED ISSUE #1085: Windows: CLI is a no-op due to path separator mismatch in isDirectExecution
## Bug

On Windows, running `exarchos mcp` (or any subcommand) silently exits with code 0 and no output. The CLI is a complete no-op.

## Root Cause

In `servers/exarchos-mcp/src/index.ts`, the `isDirectExecution` guard compares `import.meta.url` against `process.argv[1]`:

```ts
const isDirectExecution =
  process.argv[1] &&
  (import.meta.url.endsWith(process.argv[1]) ||
    import.meta.url.endsWith(process.argv[1].replace(/\.ts$/, '.js')));
```

On Windows:
- `import.meta.url` = `file:///C:/Users/.../exarchos.js` (forward slashes)
- `process.argv[1]` = `C:\Users\...\exarchos.js` (backslashes)

`endsWith` never matches, so `main()` never runs.

## Fix

Normalize backslashes before comparison:

```ts
const isDirectExecution =
  process.argv[1] &&
  (import.meta.url.endsWith(process.argv[1]) ||
    import.meta.url.endsWith(process.argv[1].replace(/\\/g, '/')) ||
    import.meta.url.endsWith(process.argv[1].replace(/\.ts$/, '.js')));
```

## Environment

- Windows 11
- Node.js v20.17.0 (also tested with v22.21.0 via Agency)
- `@lvlup-sw/exarchos@2.6.0` installed globally via `npm install -g`
- Also affects `npx @lvlup-sw/exarchos mcp`

## Workaround

Patch `dist/exarchos.js` after install to add the backslash normalization.

## Additional Note

The bundled `better-sqlite3` native binary also fails on this environment ("not a valid Win32 application"), but the JSONL fallback works correctly. This may be a separate issue related to the native binary being built for a different Node.js version/platform.

## Source

Reported in `docs/bugs/exarchos-windows-bug.md`.

rsalus:
## Summary

- Fixes #1085: on Windows the `exarchos` CLI silently exits with code 0 because `import.meta.url` (forward slashes) and `process.argv[1]` (backslashes) never match in the `endsWith` guard, so `main()` never runs.
- Extracts the inline guard into an exported `isDirectExecution(metaUrl, argv1)` that normalizes backslashes before comparing, mirroring the existing exported `isMcpServerInvocation` helper.
- Adds six unit tests pinning POSIX direct, Windows direct, `.ts → .js` tsx invocation (POSIX + Windows), library-import, and missing-argv cases.

## Changes

- `servers/exarchos-mcp/src/index.ts` — replace inline `isDirectExecution` const with exported pure function; normalize `\\` → `/` before `endsWith` comparison; add docstring referencing #1085.
- `servers/exarchos-mcp/src/index.test.ts` — add `describe('isDirectExecution (#1085)')` block with six regression tests.
- `docs/bugs/exarchos-windows-bug.md` — commit the original bug-report doc that the issue body already references as the source-of-truth.

## Test Plan

- [x] `npm run test:run` in `servers/exarchos-mcp` — 5075 passed / 21 skipped / 0 failed (19.24s)
- [x] `npm run typecheck` at repo root — clean
- [x] RED commit (`ce826549`) verified failing before GREEN — "isDirectExecution is not a function" on all six new tests
- [x] GREEN commit (`d54cd822`) turns all six new tests green
- [ ] **Windows smoke test** — cannot run from this Linux dev host; reporter (or Windows CI once added) should confirm `exarchos mcp` no longer silently exits. The related follow-up is adding Windows CI coverage so this regression category can't slip through again.

### Follow-ups (not in scope)

- Windows CI job for the MCP server package — no current Windows coverage is the reason this regression shipped in 2.6.0.
- Bundled `better-sqlite3` native binary also fails on Windows ("not a valid Win32 application"); JSONL fallback works. Separate issue — worth filing once this ships.

### DIFF
--- a/docs/bugs/exarchos-windows-bug.md
+++ b/docs/bugs/exarchos-windows-bug.md
@@ -1,3 +1,9 @@
+---
+title: Windows CLI no-op due to path separator + URL encoding mismatch
+issue: 1085
+tags: [bug, windows, cli]
+---
+
 # Windows: CLI is a no-op — isDirectExecution check fails due to path separator mismatch
 
 ## Bug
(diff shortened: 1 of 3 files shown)


rsalus:
Addressed all three CodeRabbit findings in commits [659c5440](https://github.com/lvlup-sw/exarchos/commit/659c5440) (RED) and [687c1b42](https://github.com/lvlup-sw/exarchos/commit/687c1b42) (GREEN):

**1. `fileURLToPath` for percent-encoded URLs** (Major) — adopted. `import.meta.url` is a file:// URL and percent-encodes spaces / non-ASCII. Paths like `/Users/First Last/...` or `C:\Users\John Doe\...` would have silently failed direct-execution detection even after the prior slash-normalization fix. Routing through `fileURLToPath()` decodes the URL and returns a platform path — the three `endsWith` branches collapse to two since URL and argv are now in the same normalized form. Added two regression tests (POSIX + Windows, both with percent-encoded spaces).

**2. Doc fix snippet mismatch** (Minor) — replaced the stale inline-normalization snippet in `docs/bugs/exarchos-windows-bug.md` with the real `fileURLToPath` implementation so the bug-report doc stays copy-safe.

**3. Frontmatter** (Minor) — added `title`, `issue`, `tags` frontmatter to the bug report.

### Validation
- 18/18 tests in `src/index.test.ts` green (16 from initial PR + 2 percent-encoded regression tests added as RED in 659c5440, turning GREEN in 687c1b42)
- Full MCP suite: **5077 passed / 21 skipped / 0 failed**
- `npm run typecheck` clean

RED commit for the percent-encoded cases is preserved (659c5440) so the TDD signal is visible in history — the two new tests fail on the prior normalization-only impl, which is exactly what CodeRabbit predicted.

PR #1133, opened 2026-04-18
fix(mcp): surface prune/checkpoint/agents/plugins in describe config (#1130)

### LINKED ISSUE #1130: describe(config:true): missing `prune`, `checkpoint`, `agents` sections after PR #1124
## Impact

`exarchos_workflow describe config:true` returns only `review`, `vcs`, `workflow`, `tools`, `hooks`. PR #1124 added three new config sections (`prune`, `checkpoint`, `agents`) with full schema + resolution support, but the describe surface was never extended. Users cannot inspect these values through the documented introspection path.

## Repro

With the following in `.exarchos.yml`:

```yaml
prune:
  stale-after-days: 7
  phase-exclusions: [delegate, review, synthesize, implementing]
  malformed-handling: report
  require-dry-run: true
agents:
  default-model: opus
  models:
    scaffolder: haiku
```

```
$ exarchos_workflow describe config:true
{
  "config": {
    "review": {...},
    "vcs": {...},
    "workflow": {...},
    "tools": {...},
    "hooks": {...}
    # no prune, no agents, no checkpoint
  }
}
```

But direct `resolveConfig()` against the same yml returns the correct resolved values:

```
prune.staleAfterDays: 7
agents.models.scaffolder: haiku
checkpoint.operationThreshold: 20   (default — I did not set)
```

So resolution works; describe is the only broken surface.

## Root cause

`servers/exarchos-mcp/src/workflow/describe-config.ts:23-80` — `buildConfigDescription` returns a hardcoded object literal with only the pre-#1124 sections. No per-section extension point, no iteration over `ResolvedProjectConfig` keys.

## Fix

Extend `buildConfigDescription` to emit `prune`, `checkpoint`, `agents` sections. Ideally refactor to iterate `ResolvedProjectConfig` keys with per-section annotation rules, so future sections don't require a parallel edit.

Also consider: a test that asserts every key of `ResolvedProjectConfig` appears in the describe output — would have caught this at PR #1124 review time.

## Minor related observation

`vcs.provider` annotates as `source: "default"` when yml says `provider: github` — because the default happens to be `"github"` and `annotate()` uses `JSON.stringify` equality. Not strictly a bug (the effective value is the same), but it can mislead users into thinking their explicit setting wasn't loaded. Consider tracking "raw was present" separately from "value differs from default."

## Related

- PR #1124 (prune + checkpoint + agents config landed; describe missed)
- Sibling audit finding that schema registration gaps ship to prod unnoticed (#1127, #1129)

rsalus:
## Summary

- Extend `buildConfigDescription` to emit `prune`, `checkpoint`, `agents`, and `plugins` sections — all four were present in `ResolvedProjectConfig` but silently dropped from the describe response
- Discovered scope: filed #1130 for 3 sections; fix also covers `plugins` (same shape of bug)
- Each field annotated with `source` (default vs `.exarchos.yml`) per existing pattern

## Changes

- `servers/exarchos-mcp/src/workflow/describe-config.ts` — 33 lines added, mirrors the existing section-by-section annotate pattern
- `servers/exarchos-mcp/src/workflow/describe-config.test.ts` — 9 new tests (section presence + default/override pairs per section)

## Test Plan

- [x] 9 new tests fail RED before the fix; all pass GREEN after
- [x] Full mcp suite: 5499 pass, 0 fail, 21 skipped (368 test files)
- [x] Typecheck clean

Closes #1130

### DIFF
--- a/servers/exarchos-mcp/src/workflow/describe-config.test.ts
+++ b/servers/exarchos-mcp/src/workflow/describe-config.test.ts
@@ -22,15 +22,10 @@ describe('buildConfigDescription', () => {
 
   it('DescribeConfig_AllSectionsPresent', () => {
     const result = buildConfigDescription(DEFAULTS);
-    expect(result).toHaveProperty('review');
-    expect(result).toHaveProperty('vcs');
-    expect(result).toHaveProperty('workflow');
-    expect(result).toHaveProperty('tools');
-    expect(result).toHaveProperty('hooks');
-    expect(result).toHaveProperty('prune');
-    expect(result).toHaveProperty('checkpoint');
-    expect(result).toHaveProperty('agents');
-    expect(result).toHaveProperty('plugins');
+    // Parity check — any top-level section added to DEFAULTS must appear
+    // in the description output (and vice versa), so future sections can't
+    // silently regress.
+    expect(Object.keys(result).sort()).toEqual(Object.keys(DEFAULTS).sort());
   });
 
   it('DescribeConfig_GateOverride_ShowsGateAndDimension', () => {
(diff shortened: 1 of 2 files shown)


rsalus:
@coderabbitai please re-review — the test now uses the DEFAULTS key-parity assertion you suggested.

PR #1144, opened 2026-04-18
fix: rename agent_spec.format → outputFormat + guard registration collisions (#1127)

### LINKED ISSUE #1127: MCP: `doctor` and `init` actions unreachable — format enum collision in buildRegistrationSchema
## Impact

Discovered while dogfooding v2.8.0 on `main`. The `doctor` (#1089/#1114) and `init` (#1116) orchestrate actions cannot be invoked via MCP at all — every invocation fails input validation.

## Reproduction

Both calls fail:

```jsonc
// MCP call — no format argument
exarchos_orchestrate({ action: "doctor" })
// => INVALID_INPUT: format: Expected 'table' | 'json', received 'full'

// MCP call — explicit format
exarchos_orchestrate({ action: "doctor", format: "json" })
// => MCP error -32602: format: Expected 'full' | 'prompt-only', received 'json'
```

`exarchos_orchestrate({ action: "init", nonInteractive: true })` fails with the same first error.

CLI (`exarchos doctor --format json`) works — routed through a different adapter.

## Root cause

`servers/exarchos-mcp/src/registry.ts:134-155` — `buildRegistrationSchema` iterates actions and does `if (key in shape) continue`, so the **first** action to register a `format` field wins. `agent_spec` (line 1179) is listed before `doctor` (line 1186) and `init` (line 1287), and its schema declares:

```ts
format: z.enum(['full', 'prompt-only']).default('full')  // agents/handler.ts:19
```

This default `'full'` gets injected into every orchestrate call (breaking the no-format path for doctor/init), and the enum itself refuses `'table' | 'json'` at validation time (breaking the explicit-format path).

## Why tests didn't catch it

`doctor.parity.test.ts` uses `harnessCallMcp` which calls `dispatch()` directly, bypassing the MCP SDK registration schema where the collision occurs. The parity test suite measures handler-level parity, not end-to-end MCP callability.

## Proposed fix

Disambiguate the overlapping field — three candidate approaches:

1. **Rename one enum** (smallest diff). `agent_spec` uses `outputFormat` (already present as `format: 'full' | 'prompt-only'` for prompt vs full-spec emission); doctor/init continue using `format` for table vs JSON.
2. **Merge enums** in `buildRegistrationSchema` — union all per-action enums for overlapping keys, drop defaults.
3. **Per-action schema namespacing** — nest per-action fields under `args` in the flat registration schema.

(1) is least disruptive. (2) is most future-proof but changes the public JSON Schema shape.

## Test gap

Add a real MCP-roundtrip test — invoke through the registered SDK tool, not `dispatch()` — for at least one action per composite tool, so this class of collision fails CI.

## Related

- #1089 (doctor epic), #1114 (doctor PR), #1116 (init PR) — both surfaces shipped broken on MCP.

rsalus:
## Summary

Fixes #1127 — `doctor` and `init` actions were unreachable via MCP because `agent_spec.format` (enum `full|prompt-only`, default `full`) shadowed `doctor.format` and `init.format` (enum `table|json`) in `buildRegistrationSchema`'s silent first-wins field merge. The SDK-level schema injected the wrong default into every `exarchos_orchestrate` call and rejected documented-valid payloads at the MCP boundary.

### Fix (renames the colliding field)
- `servers/exarchos-mcp/src/agents/handler.ts` — `agent_spec.format` → `agent_spec.outputFormat`
- `servers/exarchos-mcp/src/agents/handler.test.ts`, `servers/exarchos-mcp/src/orchestrate/composite.test.ts` — test call-sites updated
- `documentation/reference/tools/orchestrate.md` — docs updated with cross-reference to #1127

### Guard (prevents the class of bug)
- `servers/exarchos-mcp/src/registry.ts` — `buildRegistrationSchema` now throws at registration time when two actions declare the same field with:
  - incompatible **base types** (e.g., `string` vs `number`)
  - different **enum value sets**
  - different **defaults**

  Constraint drift (min/max/optionality) remains first-wins because handler-level schemas re-validate via `dispatch()`. Guard is scoped narrowly on purpose — it catches #1127-class hazards without flagging latent constraint-drift as errors. (The guard surfaced pre-existing drift on `stateFile`, `designPath`, and `prNumbers` across several actions; those are not reachability bugs and are left for a separate hygiene pass.)

### Regression coverage
- `servers/exarchos-mcp/src/registry.test.ts` — 6 new tests, including ones that exercise the real `exarchos_orchestrate` registry through the registration schema (not `dispatch()`), closing the DIM-4 test-fidelity gap identified in #1127.

## Test Plan

- [x] `npm run test:run` (root) — 604/604 ✓
- [x] `servers/exarchos-mcp && npm run test:run` — 5497/5497 ✓
- [x] `npm run typecheck` — ✓
- [x] New regression tests assert:
  - `exarchos_orchestrate({action:'doctor'})` parses
  - `exarchos_orchestrate({action:'doctor', format:'json'})` parses
  - `exarchos_orchestrate({action:'init', nonInteractive:true})` parses
  - `exarchos_orchestrate({action:'agent_spec', agent:'implementer', outputFormat:'full'|'prompt-only'})` parses
  - `buildRegistrationSchema` throws on synthetic enum/base-type collisions
  - `buildRegistrationSchema` accepts structurally-identical shared fields

Closes #1127.

### DIFF
--- a/documentation/reference/tools/orchestrate.md
+++ b/documentation/reference/tools/orchestrate.md
@@ -360,7 +360,7 @@ Retrieve an agent specification for subagent dispatch. Returns the agent's syste
 {
   "action": "agent_spec",
   "agent": "implementer",
-  "format": "full"
+  "outputFormat": "full"
 }
 ```
 
@@ -368,7 +368,7 @@ Retrieve an agent specification for subagent dispatch. Returns the agent's syste
 |-----------|----------|------|-------------|
 | `agent` | yes | string (enum) | Agent identifier from the registered spec list |
 | `context` | no | object | Key-value pairs for template variable interpolation in prompts |
-| `format` | no | `"full"` \| `"prompt-only"` (default: `"full"`) | `full` returns the complete spec; `prompt-only` returns just the system prompt |
+| `outputFormat` | no | `"full"` \| `"prompt-only"` (default: `"full"`) | `full` returns the complete spec; `prompt-only` returns just the system prompt. Renamed from `format` in #1127 to avoid a registration collision with the `format: "table" \| "json"` parameter on `doctor` and `init`. |
 
 Phases: all. Role: `any`.
 
--- a/servers/exarchos-mcp/src/agents/handler.test.ts
+++ b/servers/exarchos-mcp/src/agents/handler.test.ts
@@ -6,7 +6,7 @@ import { handleAgentSpec, agentSpecSchema } from './handler.js';
 describe('handleAgentSpec', () => {
   it('AgentSpec_ValidAgent_ReturnsFullSpec', async () => {
     // Arrange
-    const args = { agent: 'implementer' as const, format: 'full' as const };
+    const args = { agent: 'implementer' as const, outputFormat: 'full' as const };
 
     // Act
     const result = await handleAgentSpec(args);
@@ -62,7 +62,7 @@ describe('handleAgentSpec', () => {
         requirements: 'Must validate email format',
         filePaths: 'src/login.ts, src/login.test.ts',
       },
-      format: 'full' as const,
+      outputFormat: 'full' as const,
     };
 
     // Act
@@ -87,7 +87,7 @@ describe('handleAgentSpec', () => {
       context: {
         taskDescription: 'Build it',
       },
-      format: 'full' as const,
+      outputFormat: 'full' as const,
     };
 
     // Act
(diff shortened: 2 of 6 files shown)


rsalus:
@coderabbitai please re-review — classified ZodLiteral/NativeEnum/Union in extractEnumValues and added the default-only collision regression test you suggested.

Issue #1434, opened 2026-05-16
exarchos_view pipeline crashes on internal `__migration__` stream (validator drift)


rsalus:
## Summary

`exarchos_view {action: "pipeline"}` fails on a clean install with:

```
VIEW_ERROR: Invalid streamId: "__migration__" — must match /^[a-z0-9-]+$/
```

This makes the pipeline view (and anything that depends on it, e.g. `/exarchos:rehydrate` without an explicit featureId) unusable. Repro: fresh repo, no active workflows, call `exarchos_view pipeline`.

## Root cause

Validator drift between the storage and views layers.

| Layer | File | Pattern | Allows `__migration__`? |
|---|---|---|---|
| Shared storage validator | `servers/exarchos-mcp/src/shared/validation.ts:33` | underscores allowed | yes |
| Event-store internal sentinel | `servers/exarchos-mcp/dist/storage/jsonl-importer.d.ts:29-32`, `migration-lock.d.ts:17` | (writes events on `__migration__`) | — |
| **Views snapshot store** | `servers/exarchos-mcp/src/views/snapshot-store.ts:16` | `/^[a-z0-9-]+$/` (kebab only) | **no** |

The event store legitimately writes events to the internal `__migration__` stream (jsonl-importer + migration-lock emit on it). When `ViewMaterializer` iterates streams to snapshot/load them, `SnapshotStore.getSnapshotPath` runs `assertSafeId(streamId, 'streamId')` against the kebab-only `SAFE_ID_PATTERN` and throws — aborting the whole pipeline call.

The two validators disagree about what a legal streamId is; the views side wins by crashing.

## Fix options

- **(a)** Have views skip internal `__`-prefixed sentinel streams before snapshotting. Narrower; preserves the intentional kebab-only constraint on user-facing featureIds.
- **(b)** Align `views/snapshot-store.ts` `SAFE_ID_PATTERN` with `shared/validation.ts` to allow underscores. Sentinels are an internal contract, not user input — but this widens what views will silently materialize.

Recommend **(a)** — explicit allowlist or `__`-prefix skip in the stream iterator.

## Repro

```bash
exarchos_view {action: "pipeline"}
# → VIEW_ERROR: Invalid streamId: "__migration__" — must match /^[a-z0-9-]+$/
```

Discovered via `/exarchos:dogfood` triage during a `/exarchos:rehydrate` call with no featureId provided.

PR #1495, opened 2026-05-30
fix(views,rehydrate): state-source integrity — pipeline validator drift (#1434) + cold-probe store mutation

rsalus:
## Summary

Fixes a cluster of three related state-source-integrity defects in `servers/exarchos-mcp`, surfaced by a `/exarchos:dogfood` trace. Full RCA: [`docs/rca/2026-05-30-state-source-integrity.md`](docs/rca/2026-05-30-state-source-integrity.md).

- **CB-1 (HIGH, regresses #1434):** `exarchos_view pipeline` crashed with `VIEW_ERROR: Invalid streamId` on any store containing a two-segment slash stream id (`elicitation/<uuid>`, `workflow-state/<id>`, `workflow/<id>`, `invariants/user`). The write-side `validateStreamId` accepts these; the projection-side `SnapshotStore.SAFE_ID_PATTERN` does not. #1434 patched only `__`-prefixed ids at the materializer, so the slash classes still crashed the view.
- **CB-2 (MEDIUM):** `exarchos_workflow rehydrate` of a never-`init`'d featureId emitted a `workflow.rehydrated` event unconditionally — a pure cold probe mutated the store, materializing a phantom stream (lone audit event, no `workflow.started`, no `streams`/`workflow_state` row). A read-only audit found **14 such phantom streams** already in the local store.
- **DOC-1 (MEDIUM):** the SQLite-vs-`<id>.state.json` dual source-of-truth was undocumented, which led an agent to declare a branch "untracked" from filesystem evidence alone.

## Changes

- **CB-1** — Export a single `isSnapshotSafeId` predicate from `views/snapshot-store.ts` and filter non-snapshot-safe ids at the pipeline's **discovery** site (`views/tools.ts`), where `isFeatureStream` already filters. This is the correct layer: enumeration tolerance lives at discovery, while explicit single-id view queries keep validating their `workflowId` argument via the `materialize`/`getSnapshotPath` throw (input validation). The `materializer.ts` #1434 sentinel skip is left untouched. Converts the obsolete skipped pipeline test into an active regression guard across all id classes.
- **CB-2** — Guard the `workflow.rehydrated` emission on `streamIsEmpty = snapshot === undefined && tailEvents.length === 0`; a cold probe is now side-effect-free. Surface `_meta.workflowExists` so callers disambiguate "never existed" from "tracked but empty" without inspecting the filesystem. The documented `success:true` cold-probe contract is preserved.
- **DOC-1** — Document the two-surface model + canonical existence check (`_meta.workflowExists` / non-empty `workflowState.featureId`, never `.state.json` presence) in `CLAUDE.md` Architecture and `commands/rehydrate.md`.

## Test Plan

- [x] CB-1: converted-from-skipped pipeline regression test seeds `elicitation/*`, `workflow-state/*`, `workflow/*`, `invariants/user`, `__migration__` alongside a normal stream → `handleViewPipeline` returns success, excluding the unprojectable streams. RED confirmed before fix.
- [x] CB-1: the two `handleViewWorkflowStatus`/`handleViewTasks` "invalid characters" tests still pass — explicit-input validation is preserved (this is what flagged the wrong-layer first attempt).
- [x] CB-2: new tests — cold probe of an unknown featureId emits **no** event + `_meta.workflowExists === false`; a probe of an existing stream still emits and flags `workflowExists === true`. RED confirmed before fix.
- [x] Full MCP suite: **7190 passed / 0 failed / 21 skipped**.
- [x] `tsc --noEmit` (MCP + root) clean; `npm run skills:guard` exit 0 (no `skills/` drift).

> Follow-up (not in this PR): clean the 14 pre-existing phantom rehydrated-only streams from the local event store (data cleanup; CB-2 prevents new ones).


rsalus:
Addressed the CodeRabbit review (pullrequestreview-4395935414): added the `text` language identifier to both fenced code blocks in the RCA (the `exarchos_view` and `exarchos_workflow` examples) for MD040 compliance — pushed in 00cac8a9.

rsalus:
## Folded in: CI `npm ci` stall fix (functional, not band-aid)

The CI stall blocking this PR was the recurring MCP-server `npm ci` hang (cf. #1492 'retrigger after stuck npm ci'). Diagnosed it to the install step (all heavy jobs wedged 27 min in 'Install MCP server dependencies'); root causes were a cold npm cache + a heavy, mostly-unnecessary dev tree. Fixes (commits `40757906`, `b4bea057`):

- **`promptfoo` is eval-only** (dynamic `import('promptfoo')` in two graders; mocked in unit tests, live only under `RUN_EVALS`). The **Binary Matrix** now installs with `npm ci --omit=dev` — it needs only the 5 prod deps for `bun build --compile`. Result: matrix jobs went from *27 min hung* → **~20s green**.
- **Latent packaging bug surfaced + fixed:** `gray-matter` is imported by the compiled binary (`src/architecture/invariants-loader.ts`) but was a `devDependency` — moved to `dependencies`. Verified locally: `npm ci --omit=dev` + `bun run scripts/build-binary.ts` builds cleanly (850 modules). It's the only bundle-reachable devDep.
- **Caching:** `cache-dependency-path` now keys on the nested `servers/exarchos-mcp/package-lock.json` on every job, so the eval/test jobs install warm.
- **Thin safety net:** `scripts/npm-ci-retry.sh` wraps every `npm ci` with a per-attempt timeout + retry, so a transient registry/CDN stall (or `better-sqlite3`'s GitHub-Releases prebuild, which npm cache can't cover) costs one timeout window, not the whole job.

RCA: `docs/rca/2026-05-30-state-source-integrity.md`.

rsalus:
Addressed CodeRabbit review (pullrequestreview-4396013923) — both findings on `scripts/npm-ci-retry.sh`, pushed in `c2ffdda0`:

1. **kill-after grace** — `timeout` now uses `-k "${kill_after_s}s"`, so if `npm ci` (or a child like `node-gyp`/`prebuild-install`) ignores SIGTERM, it's SIGKILLed after the grace rather than letting `timeout` itself block on the ignorer. Valid and directly on-point for the stall we're fixing.
2. **no sleep after the final attempt** — the `sleep`+retry-log are now guarded by `if [ "$i" -lt "$attempts" ]`.

Also made attempts / timeout / kill-after / retry-sleep env-tunable, and added the requested co-located test **`scripts/npm-ci-retry.test.sh`** covering success, retry-then-success, exhaust-all-attempts (asserting exit 1 + exactly N−1 retry messages, i.e. no trailing sleep), and a SIGTERM-ignoring stall that must be SIGKILLed then recover (guards the `-k` behaviour — without it the test would hang and trip its outer `timeout`). Wired into the `grep-gates` CI job; all 5 assertions pass locally.

Issue #1696, opened 2026-07-15
Mechanize the one-copy rule: grep gate for the projection-scope claim (#1342 took 4 review cycles without it)


rsalus:
## Summary

#1342 retired the global projection scope. Its adversarial review took **four cycles**, and every cycle failed on the same defect: the scope guarantee was restated in ~8 places across 5 files, and each pass found restatements the last one missed.

The remediation consolidated the claim to two homes (`projections/types.ts`'s `scope` docstring and `docs/architecture/projections.md`'s "Reducer scope discipline"), and `types.ts` now states the one-copy rule explicitly. **But that rule is prose, and prose has no compiler.** It is enforced only by whoever reads it before writing — which is the exact enforcement model that produced #1342 in the first place.

The reviewer's framing is worth keeping: *the epic that exists because prose rotted has just shipped the machinery that stops prose from rotting, and hasn't pointed it at itself.*

## The ask

Add `scripts/check-scope-claim-single-copy.sh` — a grep gate, modeled directly on `scripts/check-begin-immediate-substrate.sh`, which #1342's task 006 just rewrote and task 007 just wired into the blocking `grep-gates` job. The pattern is established end to end; this points it at the claim.

Roughly 40 lines:
- Grep `servers/exarchos-mcp/src` + `docs/architecture` for the phrasings: `unrepresentable`, `not representable`, `cannot be authored`, `unauthorable`, `only scope`, `can carry`, `lives entirely in the type`, `type is the guard`, `cross-stream fold`.
- Allowlist the two homes: `projections/types.ts` and `docs/architecture/projections.md`.
- Allowlist sites that legitimately scope their own probe/tombstone (`types.test.ts`, `aggregate-stream.test.ts`) and the local-safety prose at `atomic-appender.ts:473/708/804`, which states site facts rather than restating the guarantee.
- Print `file:line`, exit 1 on any other hit.
- Add a self-test beside the two wired in at `ci.yml:510`, following `check-begin-immediate-substrate.test.sh`'s shape (seeded fixture fails; real tree passes; kill-probe the gate itself).

## Why it's worth doing

**It would have caught findings F, G, H plus `worktrees.ts:546` and `taskstore/types.ts:33` in one CI run instead of four review cycles.** Two of those five were found only by sweeping phrasings rather than files, and both lived in files that tasks 001 and 003 had already edited and marked done — the restatement survived *because* the file was "finished."

[...]

Issue #1695, opened 2026-07-15
package-lock.json drift: version bump e83a8f9a left lock at preview.2, so any npm install dirties the tree


rsalus:
## Summary

`servers/exarchos-mcp/package.json` is at `2.12.0-preview.3`, but `servers/exarchos-mcp/package-lock.json` still says `2.12.0-preview.2`. The version bump commit `e83a8f9a` ("chore: bump version to 2.12.0-preview.3") updated `package.json` without regenerating the lockfile.

Consequence: **any** `npm install` under `servers/exarchos-mcp` silently rewrites `package-lock.json` to sync the version, dirtying the working tree.

## Evidence

```
$ git show e83a8f9a:servers/exarchos-mcp/package.json      | grep -m1 '"version"'
  "version": "2.12.0-preview.3",
$ git show e83a8f9a:servers/exarchos-mcp/package-lock.json | grep -m1 '"version"'
  "version": "2.12.0-preview.2",
```

The resulting churn on any `npm install` is exactly two fields (the root `version` and the self-referential package entry):

```
-  "version": "2.12.0-preview.2",
+  "version": "2.12.0-preview.3",
-      "version": "2.12.0-preview.2",
+      "version": "2.12.0-preview.3",
```

## Why it matters

It is small but it taxes every agent and contributor who touches the server, and it does so in a way that invites a bad outcome:

- Nested worktrees **require** `cd servers/exarchos-mcp && npm install` before the server typecheck will run at all (otherwise it fails with misleading `Cannot find module 'pino'` / `'better-sqlite3'` errors). So this is on the mandatory path, not an edge case.
- Every one of the four #1342 implementer agents that ran `npm install` hit this independently, and each spent effort noticing it, deciding it was unrelated, and reverting it to keep their commit atomic.
- The failure mode when someone *doesn't* notice is an unrelated lockfile change riding along in a feature PR — or, worse, being reverted by hand and drifting further.

It is also self-perpetuating: because everyone reverts it, the lockfile never gets fixed, so the next person hits it too.

## Suggested direction

- Regenerate the lockfile on `main` (`cd servers/exarchos-mcp && npm install`, commit the two-line change) — the one-time fix.
- Then prevent recurrence: have the release/version-bump path run `npm install --package-lock-only` (or equivalent) so `package.json` and the lockfile move together. Worth checking whether the root package and any other workspace member have the same gap, since the bump process is shared.
[...]

Issue #1694, opened 2026-07-15
check-withsession-idempotency.sh is vacuous: selects zero production files, so the CI gate guards nothing


rsalus:
## Summary

`scripts/check-withsession-idempotency.sh` currently guards **nothing**. It selects files to scan solely by the presence of a `.withSession(` call site, and **zero production files in the repo have one**. The gate therefore selects an empty file set and exits `0` vacuously — everywhere, always.

The CI step `withSession idempotency contract (#1342 P1.D)` (`.github/workflows/ci.yml:494`) has been green since it was added, and that green means "no files matched", not "the contract holds".

This is the same defect class as the `BEGIN IMMEDIATE` literal gate fixed in #1342 task 006: a grep-gate whose selector never matches, passing vacuously while reading as protection.

## Evidence

Verified against `worktree-1342-marten-consumers` @ `7b0c7465`:

```
$ grep -rlE '\.withSession\(' servers/exarchos-mcp/src/ | grep -vc test
0
```

The selector is `find_matching_files` (`check-withsession-idempotency.sh:69`/`:81`), which matches on `'\.withSession\('` alone. No call site means the file is never selected, so the gate can neither pass nor fail it.

Independently corroborated from the other direction by #1342 task 009: `src/orchestrate/vcs/create-pr.ts` carried a comment claiming it "satisfies the CI idempotency contract gate". It does not — it has no `.withSession(` call and appends directly, so the gate never scans it. That comment has been corrected on the #1342 branch.

## Why it matters

`withSession` is dormant-and-correct: it is available for the right shape (Marten `FetchForWriting` posture) and has intentionally zero consumers today. That is fine. What is not fine is a gate that reports success over an empty set, because:

1. It reads as evidence the idempotency contract is enforced. It is not.
2. The moment a first `withSession` consumer lands, the gate starts scanning with no track record — nobody has ever seen it fail on real code.
3. It cost review time on #1342 twice: once via the `create-pr.ts` comment, once here.

## Suggested direction (not prescriptive)

Options worth weighing, in rough order of appeal:

- **Fail loudly on an empty selection.** If the gate selects zero files, exit non-zero with "selector matched no files — this gate is not guarding anything". Converts a silent vacuity into a visible one. Cheapest, and it generalizes to every grep-gate.
- **Assert the expected consumer set.** Pin the count/paths the gate expects to scan, so a drop to zero is a failure rather than a pass.
[...]

Issue #1643, opened 2026-07-04
DR-2: full re-invocation idempotency for prepare_review scope:plan (client token)


rsalus:
## Summary

`prepare_review scope:plan` (the DR-2 revision count+cap seam in `servers/exarchos-mcp/src/orchestrate/prepare-review.ts`, `buildPlanReviewProvisioning`) counts prior dispatches, checks the cap, then appends a `workflow.plan-review-dispatched` marker keyed by ordinal. The read→check→append is **not atomic**, so two concurrent same-`featureId` re-dispatches can both read `priorDispatches.length = N`, both pass the cap, and both append under the same `${featureId}:plan-review-dispatch:${N}` idempotency key — the INV-8 UNIQUE index collapses one append while both calls return a provisioning payload. Net: a benign **under-count** (the cap becomes slightly more lenient), never a fail-open safety breach.

Surfaced by CodeRabbit on PR #1642 (Major, data-integrity) and consciously deferred there.

## Why it was deferred (not a live bug today)

- **Unreachable in the governed flow.** `prepare_review scope:plan` is driven by a single orchestrator inside the HSM-guarded `plan → plan-review → plan` loop (INV-9); a feature can't be in `plan-review` twice concurrently, and the seam is issued sequentially. The only genuine window is a crash-and-retry, already documented in the `buildPlanReviewProvisioning` docstring.
- **The append is INV-8 idempotency-keyed by design** — same-ordinal re-append dedups at the storage layer; that keyed dedup is the deliberate boundary, not full handler-invocation idempotency.
- **Naive OCC is the wrong tool here.** CAS-pinning an `expectedSequence` onto an idempotency-keyed append hits a known trap in this codebase: the idempotency cache-hit precedes the CAS check and wedges the retry loop.

## Proposed fix (when prioritized)

Full re-invocation idempotency via a **client-supplied token** (out of DR-2 scope), so each provisioning maps 1:1 to a durable dispatch marker regardless of concurrency or crash-retry. Add a regression test for the concurrent double-dispatch path.

## References

- PR #1642 (DR-2 seam) · epic #1574 (WLM, closed) · INV-7 / INV-8 / INV-9
- In-code follow-up note: `buildPlanReviewProvisioning` docstring in `prepare-review.ts`

## Test Plan

- Concurrent same-`featureId` `prepare_review scope:plan` dispatches each map to a distinct durable `workflow.plan-review-dispatched` marker (no collapse-with-double-provision).
[...]

PR #1697, opened 2026-07-15
Retire the global projection path: task-store@v1's scope was wrong, not underused

### LINKED ISSUE #1342: epic: post-preview.2 leverage — wire remaining Marten primitive consumers
## Summary

v2.10.0-preview.2 (#1312) lands the Marten primitives + reference consumers. This epic tracks the follow-on work to fully wire and leverage the new substrate — the consumers preview.2 ships with are deliberately scoped (one merge-orchestrate migration via `decide`, no `withSession` consumers, deferred reader-side `workflow_type` filtering).

Without this rollout, four primitives ship as available-but-underused:
- `withSession` — zero in-tree consumers (intentional, but should be audited)
- `aggregateStream` — one consumer (Wave 4 Phase B short-circuit)
- Two-event split pattern — applied to merge-orchestrate only; other non-idempotent side-effect handlers remain at-risk for retry-storm replay
- `StorageBusyError` propagation — typed error exists but operator-facing surfaces don't distinguish it from `CONCURRENCY_CONFLICT`

## Children (P1, near-term)

- [ ] `withSession` consumer audit — find handlers that genuinely need an imperative escape hatch (peek at aggregate state between multiple `append` calls within one transaction), OR document the absence. Marten-style "available for the right shape; not the default" is OK; silent "no one knows it exists" is not.
- [ ] Two-event split rollout to remaining non-idempotent side-effect handlers: `gh pr create`, `gh pr comment`, `gh issue create`, branch deletion (cleanup), worktree removal. Same shape as Wave 4 merge-orchestrate (`*.requested` event commits before side-effect; `*.executed` event commits after). Each handler today is at risk of retry-storm replay if its surrounding `withStateRetry` boundary catches a `ConcurrencyError` or `StorageBusyError`.
- [ ] `StorageBusyError` propagation audit — verify the CLI error formatter + MCP envelope serializer surface `STORAGE_BUSY` distinctly from `CONCURRENCY_CONFLICT`; document the operator remediation for each (substrate contention vs OCC loss have different fix paths).
- [ ] CI grep gate: forbid `withSession` calls that omit BOTH `operationId` and `allowNonIdempotent: true` — catches the audit §F1.1 anti-pattern at compile-review rather than runtime.
- [ ] CI grep gate: forbid `BEGIN IMMEDIATE` outside the substrate (atomic-appender owns it; any other site is an INV-1 violation in flight).

## Children (P2, next preview)

- [ ] `decide` rollout — audit `task_complete`, `task_fail`, `synthesize`, `cleanup` orchestrate handlers for load → decide → append patterns; migrate where OCC-loss replay safety justifies the refactor.
- [ ] `aggregateStream` adoption — replace ad-hoc `query(streamId).filter(...).reduce(...)` patterns where a registered reducer would carry the same logic and add OCC discipline downstream.
- [ ] In-memory store audit — TaskStore + mergeOrchestrator move to projections in preview.2; identify the remaining in-memory stores (worktree cache, capability resolver, eval flywheel state, any `Map`-backed singletons in `core/`) and either project them or document why projection is wrong-shape for that surface.
- [ ] Snapshot cadence tuning for `task-store@v1` global projection — preview.2 defers to existing `projections/cadence.ts` machinery; verify the policy fits high-volume task events (re-snapshot threshold, cold-fold cost ceiling).

## Children (P3, deferred / cross-ref)

- [ ] `workflow_type` reader wiring — already in #1090 scope (v2.12 lifecycle verbs: `exarchos_view({action: 'ps', workflowType: ...})`). Track here, deliver there.
- [ ] Reducer-scope discipline docs + new-reducer template (`docs/architecture/projections.md`) — every new reducer must declare `scope: 'stream' | 'global'`; the API will reject misuse at runtime, but a written template prevents the failure-then-fix loop.
- [ ] Two-event split as a documented architecture pattern (`docs/architecture/runtime.md` or new `docs/architecture/process-manager.md`). Cross-link Wolverine outbox / Akka `Effect.thenRun` / Axon saga references from audit §"Cross-framework comparison".
- [ ] `decide` vs `withSession` decision tree in skill references (so future authors don't reach for `withSession` when `decide` would do — the natural pull is toward the imperative API).

## Why this is one epic, not N issues

Each item is small (<1 day, often hours). The collective shape — "finish what preview.2 started" — is what makes them coherent. Splitting prematurely loses the architectural narrative; bundling all into preview.2 itself risks the bundle never shipping. This epic is the parking lot for "the rest of the leverage" so the bundle can land at its current scope.

## INV / DIM alignment

| Concern | Coverage |
|---|---|
| **INV-1 event-sourcing integrity** | Two-event split rollout extends INV-1's safe-by-construction posture beyond merge-orchestrate to every non-idempotent side-effect handler. In-memory store audit closes the last remaining stores-as-side-databases. |
| **INV-2 facade equivalence** | `StorageBusyError` propagation audit verifies CLI and MCP carriers surface the typed error identically. |
| **INV-5b output contract** | `withSession` idempotency-contract gate documentation + grep gate ensures the `INVALID_SESSION_OPTIONS` envelope's `suggestedFix` machinery is reachable in normal usage, not a dead-letter. |
| **DIM-4 test fidelity** | Per-handler two-event-split migration adds the "ConcurrencyError-doesn't-refire-side-effect" fixture pattern from Wave 4's `PR-API-non-refire fixture`. |
| **DIM-7 resilience** | `decide` rollout extends bounded-retry safety to handlers currently using ad-hoc `appendComputed` patterns. |

## References

- Parent epic: #1312 — Marten primitives
- Preview.2 design: [`docs/designs/2026-05-10-v2-10-0-preview-2-marten-primitives.md`](https://github.com/lvlup-sw/exarchos/blob/main/docs/designs/2026-05-10-v2-10-0-preview-2-marten-primitives.md) §"What's *not* immediately leveraged"
- Audit findings: [`docs/research/2026-05-10-v2-10-pre2-implementation-audit-findings.md`](https://github.com/lvlup-sw/exarchos/blob/main/docs/research/2026-05-10-v2-10-pre2-implementation-audit-findings.md) §F1.1, §F1.2, §F2.1
- Marten lessons: [`docs/research/2026-05-08-marten-event-store-lessons.md`](https://github.com/lvlup-sw/exarchos/blob/main/docs/research/2026-05-08-marten-event-store-lessons.md)
- v2.12 lifecycle verbs (workflow_type reader): #1090, #1316

### LINKED ISSUE #1698: Windows concurrent-rename EPERM: fix the class, not just the one site with a test
## Problem

`writeStateFile` was failing on Windows with `EPERM: operation not permitted, rename` whenever two writers published concurrently. Fixed in #1697 (`dc202dbf`) by retrying the publish with a **jittered** bounded backoff, win32-only.

The **defect class is not fixed** — only the one site that had a test.

POSIX `rename(2)` guarantees an atomic replace where the loser simply overwrites. Windows does not: while a replace is in flight the destination is briefly held open, and a concurrent `MoveFileEx` fails `EPERM`/`EACCES` even though nothing is wrong. Every tmp+rename publish in the server inherits this, and **only** the one with a concurrency test is red. The rest are latent by the same logic this epic was built on: *a surface with no consumer exercising it is not thereby correct.*

This is not speculative. The same Windows CI run logs the identical error from a **second** site, where it is swallowed rather than asserted:

```
{"level":50,"subsystem":"views","err":"EPERM: operation not permitted, rename '...telemetry.telemetry.snapshot.json.tmp.1784151143392' -> '...telemetry.telemetry.snapshot.json'","msg":"Snapshot save failed"}
```

## Sites sharing the class

All publish via tmp+rename with no concurrency test:

- `src/views/snapshot-store.ts:131` — **confirmed failing in CI logs**, error swallowed into a `Snapshot save failed` warn
- `src/utils/atomic-write.ts:56` — its own docstring says "Rename is atomic on POSIX", which names the gap exactly
- `src/storage/lifecycle.ts:161`
- `src/sync/sync-state.ts:76`
- `src/orchestrate/init/writers/claude-code.ts:50`
- `src/storage/sidecar-scheduler.ts:154,166`

## Proposed shape

Extract the publish-with-retry from `workflow/state-store.ts` into one shared helper (`utils/atomic-write.ts` is the natural home, since it already owns the tmp+rename idiom and already documents the POSIX assumption) and route every site through it. One copy, per the scope-discipline rule #1697 establishes — not six.

The jitter is load-bearing and must survive the extraction: contending writers are woken by the same collision, so a fixed or purely exponential delay retries them in lockstep and they collide again. A first cut of the #1697 fix used a deterministic `5 * attempt` backoff and left one writer still failing; randomising each sleep is what actually broke the convoy.

Keep the bound. A genuine permission fault (read-only file, ACL, antivirus handle) reports `EPERM` too and is indistinguishable at this layer, so the loop must rethrow rather than hang.

## Why it was scoped out of #1697

That branch's only adversarial gate had already failed three times; sweeping six more files after the review passed would have shipped unreviewed changes on the strength of "it's the same fix". Filing instead.

## Acceptance

- One shared publish helper; no site open-codes `fs.rename` for a tmp+rename publish.
- A concurrency test per routed site — the class stays latent precisely because nothing exercises it. Do **not** satisfy this with a win32 skip: a test skipped on the only platform that can see the bug is the vacuous-gate defect of #1694.
- `views/snapshot-store.ts` no longer logs `Snapshot save failed` under the Windows lane's concurrent load.

rsalus:
## Summary

The epic asked why `task-store@v1`'s global projection is underused. **The premise inverts: it is not underused, it is wrong** — and adopting it would have corrupted state. So the deliverable is a retirement, not a consumer.

`TaskStoreState.tasks` is `Record<taskId, TaskRecord>` keyed by a bare per-feature ordinal (`'001'`), and `TaskRecord` carries no `featureId`. A cross-stream fold therefore merges feature-A's task `001` into feature-B's, and `upsertTask` clobbers via `{...prior, ...overlay}`. `scope: 'global'` did exactly one thing: it made `readProjection('task-store@v1')` *legal to call*, returning silently corrupted state to the first consumer that adopted it. `readProjection` and `projections/cadence.ts` had **zero production callers**; both real consumers already fold `.apply` per-stream. That is why preview.2's planned consumer refactor never landed — it could not have worked.

The fix is the type system, not discipline: narrowing `ProjectionScope` to `'stream'` turns the remaining sites into compile errors, so the retirement is **atomic by construction** and the corrupting configuration is unauthorable in typechecked code.

> ### On the test numbers — this was wrong in the first version of this description
>
> An earlier draft warned that CI would show 26 known-red failures and asked reviewers to discount them. **That was inaccurate and is retracted.** CI's `Root Package` job **passes** (3m0s), as does `Exarchos MCP Server`. The ~26 merge-orchestrate / store.race failures are a **local-Linux-only** phenomenon — the Windows CI lane runs the same suite and reports them green.
>
> The framing came from four agents independently confirming the 26 against a pristine *local* base, which was true locally and then got carried into a CI-facing claim where it is false. **CI is the stricter signal on this branch, not the looser one** — so a red lane here is real until proven otherwise. That mattered: the Windows lane went red on a genuine bug while those 26 were unrelated noise.
>
> Current local suite: **26 failed / 9070 passed / 25 skipped**, all 26 in the same 9 known-red files. No test regressed in this branch.

## Changes

- **Retire the global projection path atomically** (`task 001`, high tier) — delete `readProjection` + `projections/cadence.ts`, narrow `ProjectionScope` from `'stream' | 'global'` to `'stream'`, correct `taskStoreReducer.scope`. Surviving guards (view↔reducer parity, the `UnknownProjectionIdError` positive-throw) are **re-pointed, not deleted**.
- **Pin the snapshot round-trip** (`task 002`) so rehydration checkpointing is proven unchanged by the retirement.
- **Retire the stale references** (`tasks 003`, `004`, `009`) that described `task-store@v1` as global or `readProjection` as shipped, and document the reducer-scope rule in `docs/architecture/projections.md`.
- **State `handleSet`'s idempotency contract truthfully** (`task 005`) — the key carries sorted field *names*, never patch values, so the real contract is one event per `(featureId, base-version, field-name-set)`. Token work stays deferred to #1643.
- **Make the `BEGIN IMMEDIATE` gate guard the primitive that exists** (`task 006`) — it guarded a literal that never issues as SQL; the real primitive is `.immediate()`.
- **Run the two grep-gate self-tests that never ran** (`task 007`) — a gate whose self-test never runs rots into a no-op silently.
- **Bound the `writeStateFile` temp filename without breaking the sweep** (`task 008`) — the counter goes *before* the pid so the end-anchored group can only ever land on the pid.

### Windows concurrent-rename EPERM — found by this PR's own new tests (#1698)

The Windows lane went red on the two concurrency tests task 008 added: `EPERM` on concurrent `rename`. **The tests are right; the platform is the problem.** POSIX `rename(2)` defines a concurrent replace as atomic with the loser overwriting. Windows briefly holds the destination open and fails a concurrent `MoveFileEx` with `EPERM` — so identical, correct code is green on Linux and red on win32. The defect is older than this PR; task 008 merely wrote the first concurrency test for that path, and only the win32 lane can see it.

Fixed at the publish step with a bounded, **jittered**, win32-only retry — not a `skipIf(win32)`, which would be the vacuous-gate defect of #1694. The jitter is load-bearing: the first cut used a deterministic `5 * attempt` backoff and took the lane from 2 failing tests (9 and 5 rejected writes) to 1 test with 1 rejection, then stalled. Contending writers are woken by the same collision, so a fixed *or purely exponential* delay retries them in lockstep and they collide again. A convoy retrying in lockstep is still a convoy.

**#1698 is closed here rather than deferred**: ~8 other publishes shared the identical latent race with no test, which is the reasoning this epic exists to reject. One home (`utils/atomic-write.ts`) now owns the publish; `state-store`, `snapshot-store`, `lifecycle`, `sync-state`, `outbox`, `claude-code`, `mcp-json-writer`, and `onboard/hooks` all route through it, and `state-store`'s local copy is **deleted** — one copy is the point. `sidecar-scheduler` is deliberately *not* routed and says why in place: its target is unique per drainer, so the race cannot arise.

Two real defects surfaced while routing, neither visible to a green typecheck:

- **The default reached around the caller's module seam.** It defaulted to `node:fs`'s `fs.promises.rename` while every async caller imports `node:fs/promises` — same syscall, different module reference. Routing `snapshot-store` through it bypassed that file's crash-injection mock, and the test asserting "a failed rename leaves the previous snapshot intact" published the crashed payload instead.
- **The jitter test was vacuous.** It asserted delays differ *across attempts*, which a deterministic backoff satisfies — the no-jitter mutant survived it. The real property is cross-*writer*. Rewritten; it now kills the mutant.

### Scope discipline this PR establishes

The guarantee is now stated in **exactly two places** (the `scope` docstring in `projections/types.ts` and `projections.md`); every other site points at them and asserts nothing. That is the epic's own thesis applied to its own remediation: #1342 was *caused* by one claim being restated in ~8 places until the copies outlived the code and contradicted each other. Prose has no compiler, so the only defence is to have one copy. #1696 tracks turning that rule into a grep gate, since as prose it will rot by its own thesis.

## Test Plan

- [x] `npm run typecheck` — exit 0
- [x] `cd servers/exarchos-mcp && npm run typecheck` — exit 0 (**the root typecheck does not cover the server**)
- [x] Full local suite: 26 failed / 9070 passed / 25 skipped — the 26 are the pre-existing local-only baseline, all in the same 9 files. **CI itself is fully green.**
- [x] `post_delegation_check` 4/4
- [x] **The core property, proven and never refuted across 4 adversarial review cycles:** authoring `scope: 'global'` now yields TS2322 in typechecked code
- [x] Orphan sweep pinned in both directions — a dead pid reaps; a live-pid-colliding counter does not strand the file
- [x] Both grep gates now self-test in CI (21/21, incl. 4 new cases for the bypasses below)
- [x] `publishTempFile` kill-probed — 4 mutants, all die: linear backoff (2 fail), constant backoff (1), win32 gate removed (1), retry removed (4)
- [x] The win32 retry is tested **on Linux** via platform stub + injected rename — not skipped
- [x] `resolveMaxRecords` kill-probed — the old lax parser fails 4 of 6
- [x] Rebased onto `main`; 0 commits behind

### Review posture — worth knowing

Rev.3 shipped **un-paneled by explicit operator decision** (rev.1 and rev.2 were each refuted 3/3). That made code review the **only** adversarial gate on this work, and it **failed 3 times before passing on cycle 4**. Cycles 1–3 were all correct and all found execution defects rather than design defects; the core thesis was never refuted. The sharpest finding: a task's justification for deleting the runtime scope check was **fabricated but persuasive**, agreed with the conclusion already wanted, and was endorsed as exemplary before anyone checked it — the exact failure mode this epic exists to punish. The code was right; only its stated reason was false.

The process finding, ranked above every code finding: **a file list cannot see a claim.** The conflict check ran against complete file lists and still shipped a "global reducer" header 215 lines above the `scope: 'stream'` the same commit wrote. Decomposition by file is right for merge conflicts and wrong for semantic invariants.

## Follow-ups filed

- #1694 — `check-withsession-idempotency.sh` is vacuous repo-wide (selects zero files; same class as the bug task 006 fixed)
- #1696 — turn the one-copy rule into a grep gate + self-test. **This PR is the argument for it**: the one-copy rule was stated as prose and was already false in 6 places, including in the file asserting it. Prose rots by its own thesis.

Resolved during this PR: **#1695** (lockfile drift — fixed on `main`; `npm install` no longer dirties the tree).

Closes #1342
Closes #1698


rsalus:
### Windows lane: real bug, found by this PR's own new tests

The first CI run went red on `Windows Unit (MCP)` — the two concurrency tests task 008 added (`WriteStateFile_ConcurrentInProcessWriters_NeverCollideOnTempPath`, `WriteStateFile_ConcurrentWriters_NeitherObservesPartialFile`), with `EPERM: operation not permitted, rename`. Recording the triage, because the framing matters.

**The tests are right; the platform is the problem.** Giving each writer its own temp path stopped writers clobbering each other's temp file — but every writer still replaces **one** destination. POSIX `rename(2)` defines that case: atomic, loser overwrites. Windows does not — while a replace is in flight the destination is briefly held open, and a concurrent `MoveFileEx` fails `EPERM` even though nothing is wrong. Same correct code, green on Linux, red on win32.

**The defect is older than this PR; the test is new.** Both tests are absent on `main`. Concurrent `rename` onto one destination has always been EPERM-prone here — task 008 simply wrote the first concurrency test for this path, and only the win32 lane can see it. That is the test doing its job on its first run.

**Fixed at the publish step** (`eb652c8c`): retry the rename, bounded, win32-only. Safe because what is retried is the *publish*, not the write — the payload is already fully on disk and each rename stays atomic, so a reader sees old bytes or new bytes, never a torn mix. Bounded because a genuine permission fault (read-only file, ACL, antivirus handle) also reports `EPERM` and is indistinguishable at this layer, so it must rethrow rather than spin.

**Not skipped on win32.** A test skipped on the one platform that can see the bug is precisely the vacuous-gate defect this branch filed #1694 about.

Also worth correcting from the PR description: **`Root Package` passes in CI.** The 26 known-red failures cited there are a local-only phenomenon, not a CI one — so CI is the stricter signal here, not the looser one.

Scope note: ~6 other `tmp`+`rename` sites in the server share this latent Windows race (`utils/atomic-write.ts`, `views/snapshot-store.ts`, `storage/lifecycle.ts`, `sync/sync-state.ts`, `orchestrate/init/writers/claude-code.ts`, `storage/sidecar-scheduler.ts`). None has a concurrency test, so none is red. Deliberately **not** swept here — that is a follow-up, not scope creep on a branch whose only adversarial gate already failed three times.

rsalus:
### CI is green — the Windows fix landed, and the second cut is the interesting one

`Windows Unit (MCP)` **passes** (8m12s). All checks green, 0 failures.

The first cut (`eb652c8c`) was right in shape and wrong in timing: it took the lane from **2 failing tests (9 and 5 rejected writes)** to **1 failing test with 1 rejected write**, then stalled there. The premise was fine; the backoff was the bug. `5 * attempt` is deterministic, and every contending writer is woken by the *same* collision — so they all slept the same 5ms, all retried on the same tick, and collided again. A convoy retrying in lockstep is still a convoy. `dc202dbf` randomises each sleep, which is what actually breaks it; exponential alone would not have.

Worth stating plainly: **the partial fix was the useful signal.** 9+5 → 1 said the mechanism was understood and the timing wasn't, which is what pointed at lockstep rather than at the premise.

`Windows Unit (Root)` also failed once in between, on `AssertExarchosOnPath_BinaryResolvable_DoesNotThrow` — "Test timed out in 5000ms". That one is the known vitest spawn-timeout flake, not this branch: it passed, failed, then passed again on identical inputs, and these commits touch only the spec and the MCP server's `state-store.ts`, neither of which is on the root package's PATH-resolution path.

**Filed #1698** for the defect class rather than sweeping it here: ~6 other tmp+rename sites share the identical Windows race, and one of them (`views/snapshot-store.ts`) is *already logging the same EPERM in this very CI run* — swallowed into a `Snapshot save failed` warn instead of asserted. They are latent for exactly the reason this epic exists: nothing exercises them, and a surface with no consumer exercising it is not thereby correct. Sweeping six more files after the review had already passed would have shipped unreviewed changes on the strength of "it's the same fix".

Note for merge: the branch is currently **4 commits behind main**.

rsalus:
### CodeRabbit review: 5 of 6 fixed, 1 rejected with proof

Each finding was independently verified against the code before acting — including trying to refute it. CodeRabbit scored 4.5/6, but **its severity calibration was inverted on the two that matter most**: the one finding that breaks this PR's thesis was filed as a 🔵 Trivial nitpick, and its one outright false positive was filed as 🔴 Critical.

| # | Finding | Verdict | Action |
|---|---|---|---|
| `types.ts:108` | one-copy claim inaccurate | **TRUE — and it is the thesis** | Fixed `1e8180ca` |
| `check-...sh:100` | grep fail-open | **TRUE** — reproduced | Fixed `6a335010` |
| `check-...sh:77` | inline `/* */` bypass | **TRUE** — reproduced | Fixed `6a335010` |
| `snapshot-retention.ts:32` | parseInt comment | **TRUE, understated** | Fixed `41ada045` |
| `state-store.test.ts:1096` | ENOENT swallowed | **TRUE** | Fixed `41ada045` |
| `rehydrate.test.ts:1286` | sequence divergence | **FALSE** | **Rejected — see below** |

#### The nitpick was the most important finding in the review

`types.ts` claimed it and `projections.md` were "THE ONLY TWO PLACES THIS GUARANTEE IS STATED. Every other site points here and asserts nothing." **Six other sites stated it.** `atomic-appender.ts:845` reproduced all three safety points *and* the tsconfig limit and then linked here — as though a pointer appended to a restatement were a pointer. Same in `decide` and `aggregateStream`; plus `types.test.ts`, `aggregate-stream.test.ts`, `taskstore/types.ts`, `taskstore/reducer.ts`, `reducer.test.ts`, `task-detail-view.test.ts`.

That is **~8 stating sites — the exact number that docstring cites as the disease that caused #1342**. The rule was refuted by the file asserting it, and the adversarial review passed it. Deleted rather than edited, per the review's own converging move. The claim is now narrowed to what is verifiable and admits it is review-enforced only, with #1696 tracking the gate.

#### Rejecting `rehydrate.test.ts:1286` — the suggested diff would break a green test

The claim is that `projectionSequence` should equal `written.sequence`, making `toBeLessThan` fail. **The test passes, and for exactly the intended reason.**

`projectionSequence` is a **count of handled events**; `written.sequence` is a **stream coordinate**. `rehydrate.ts:190-195` says so explicitly. Trace: `workflow.started`(1) → `task.assigned`(2) → `task.completed`(3) → `task.assigned`(4) → `workflow.checkpoint`(5). The checkpoint carries no handoff sub-object, so `applyWorkflowCheckpoint` (`rehydration/reducer.ts:615-624`) returns identity without bumping the count. Count stays **4**; `written.sequence` is **5**. `4 < 5` holds.

The error is a double conflation: reading a `case` label as "handled" when that branch folds as a no-op, and comparing a count against a coordinate as if they shared units. The committable suggestion asserts `4 === 5` — it would fail, and it would **re-introduce the precise count-vs-coordinate conflation that commit `9d8a13ed` added this assertion to pin**.

One fair point survives underneath it: the assertion is quietly coupled to the checkpoint carrying no handoff payload. Noted, not changed — pinning it harder is a separate concern from the change under review.

#### Fixes that needed a second pass — worth recording

- **My first fail-closed fix didn't work, and the probe caught it.** `mapfile -t x < <(grep …)` reports *mapfile's* status, not grep's, so `|| status=$?` always saw 0 and the guard never fired. Status now comes via a temp file.
- **Rewriting the comment logic introduced two regressions**, both caught before commit: dropping the leading-`*` rule broke every JSDoc continuation line (two false positives in the real tree — and the docstring claimed a rule the code never implemented), and cutting `#` mid-line would have eaten TypeScript private fields, hiding `this.#db.immediate()` from the scan entirely.
- **On `snapshot-retention`, CodeRabbit understated its own finding.** `"10junk"→10` and `"1.5"→1` only *tighten* the cap. The case it missed is the dangerous one: `"999999999999999999999"` parses to `1e21` — finite and positive, so it sailed through as a cap of one sextillion, which *is* the "no limit" the docstring swears is unreachable. Fixed the parser, not the comment.

All four gate fixes and the parser are kill-probed: restoring the prefix rule, the fail-open grep, or the lax parser each fails a test.

rsalus:
### Second pass: all three findings were real, all three fixed (`d3ee277f`)

CodeRabbit re-reviewed the fixes and found three more. Each verified before acting; none rejected.

**Temp leak (MEDIUM) — real, pre-existing, and wider than reported.** A failed publish orphaned its staged temp next to the target. `state-store` and `atomicWriteFile` each hand-rolled the cleanup; `snapshot-store`, `lifecycle`, `sync-state` and `outbox` silently leaked on every failure. Fixed **in the publish**, not per-caller — fixing it at eight call sites would have re-created exactly the duplication this branch just deleted. The seam is now an io object (`rename` required, `unlink` optional): plain callers get cleanup from the default, and an injected fs that cannot delete (`McpJsonWriterFs` has no `unlink`) gets none, exactly as before. No interface widening, no behaviour change for those.

**My budget assertion was loose — a fair hit.** `toBeLessThan(2000)` against a *randomly jittered* total means an implementation busting the documented ~1s budget would pass most runs and fail rarely: a real defect that reads as a flake. `Math.random` is now pinned to its maximum so the test measures the worst case and asserts the actual ceiling. Kill-probed — raising the retry limit past the budget now fails it.

**The self-test fixture contradicted itself.** Test 7's fixture claimed the literal `BEGIN IMMEDIATE` never appears while its own comment contained it. Had the comment exemption broken, the *literal* scan — not `.immediate()` detection — would have produced the expected failure, so the test would pass for the wrong reason. Removed.

Three new tests pin the cleanup (including that a failing `unlink` never masks the original error, and that an `unlink`-less io still works). Kill-probed: removing the cleanup fails 2, blowing the budget fails 2.

---

### State

- **Rebased onto `main`** — 0 commits behind.
- **#1698 closed in this PR** rather than deferred.
- CI green as of `41ada045`: `Windows Unit (MCP)` ✅, `Windows Unit (Root)` ✅, `Root Package` ✅, `Grep Gates` ✅.
- `Windows Unit (Root)` flaked twice en route and passed on rerun with no code change. Across four runs it went success → failure → success → failure, and this branch's diff never touches `benchmarks/` or `test/setup/`, where all 7 failures sat (all spawn timeouts: 30015ms, 5006ms, 5002ms…). That is the known vitest spawn-timeout flake class on the Windows runner, not this branch.
- Local suite: 26 failed / 9073 passed, all 26 in the same 9 known-red merge-orchestrate/store.race files. **Those 26 are local-Linux-only — CI runs the same suite green**, which is why the original "discount the known-red" framing in this description was retracted.

The one rejected finding (`rehydrate.test.ts:1286`) stands rejected — its suggested diff asserts `4 === 5` and would re-introduce the count-vs-coordinate conflation that `9d8a13ed` added the assertion to pin.

rsalus:
@coderabbitai review

rsalus:
@coderabbitai approve

Issue #1764, opened 2026-08-09
Epic: internal-mechanics residue — overhaul follow-ons + 2026-08-08 review remainder (071-090)


rsalus:
## Summary

Tracking issue for the **internal-mechanics residue spec** — the follow-on work from the overhaul wave plus the MEDIUM/LOW remainder of the 2026-08-08 adversarial review.

Two provenances:

- **071–077** — the overhaul wave's own residue, authored when Wave 1 exited.
- **078–090** — added by the 2026-08-08 review of `internal-mechanics-overhaul`. Its **nine HIGH findings were fixed inline** on that branch (shipped in #1755, each kill-probed); these are the remainder, grouped by defect class rather than transcribed one finding per issue.

## Design requirements

| DR | Claim |
|---|---|
| DR-1 | One validator across both event write paths |
| DR-2 | One lexer port, three remaining consumers |
| DR-3 | Field-scoped amendment writes field-scoped diffs |
| DR-4 | Entrypoint predicates test identity, not filename |
| DR-5 | One authority decides event-name well-formedness |
| DR-6 | One waiver-ledger module |
| DR-7 | A gate's verdict is the gate's own verdict |
| DR-8 | A guard's scan root and its floor are both part of its claim |
| DR-9 | The guard inventory's denominator must include the guards |

## Already filed separately

- **088** → #1756 (`prepare_synthesis` measures ambient cwd)
- **089** → #1757 (mutation runner never runs where its config lives)
- **090** → #1758 (`transition` silently ignores `dryRun`)

## Sequencing

Per the spec: **078 and 085 carry the silent-wrong-answer defects** (a skipped gate reading as evidence; a failure envelope exiting 0; a short write promoted over good data). Take those first — the rest are guards that under-report, which is a slower harm than a gate that misreports.

## Two exit criteria the review added, because both were previously satisfiable while false

- `npm run validate` exiting 0 is **not sufficient on its own** — no step may report a non-pass verdict that the aggregate records as PASS (DR-7). Read the step verdicts, not just the exit code. *(This is live today: `measured-premises` prints `VERDICT: GAPS` with 11 of 13 proof rungs unprobed while the aggregate records 9/9 PASS — see task 078 item 2.)*
- **No guard may be green because of what it does not look at** (DR-8). For every guard the spec touches, the scan root and the denominator are stated in its own test, and a narrowed root fails.

## Note on 080

[...]

PR #1809, opened 2026-08-10
refactor(1764): wave 5 — one authority for event-name well-formedness (075)

rsalus:
## Summary

Wave 5 of epic #1764 — the final task, and the last of 21. Stacked on Wave 4 (`feat/imo-residue-wave4`, PR #1808). Shipped alone with its own PR and a migration note, as the spec's D1 requires: this is the only task in the programme that changes a **public runtime seam**.

**Task 075 (DR-5, HIGH):** two authorities decided what an event name may be, and they disagreed in both directions. `EVENT_NAME_PATTERN` in `schemas.ts` rejected 25 of its own built-ins — invisible because `registerEventType` applied it only to *custom* registrations while the built-ins were a literal array never fed through it. **A validator its own authoritative corpus fails, never pointed at that corpus.** The DR-3 grammar accepted all 171 but was narrower elsewhere (no digits, single-word namespaces).

The grammar survives as the single authority. `EVENT_NAME_PATTERN` is no longer authored — it is **built** from the grammar's own vocabulary and re-exported under its old name. That binding was kept deliberately rather than deleted: the census's `shippedPattern` seam is the ratchet that catches anyone re-authoring the regex by hand, so deleting it would have deleted the instrument protecting the repair.

## Changes

- `event-name.ts` — `assertWellFormedEventName` (throws carrying the violated clause **and** the migration path), `buildEventNamePattern` deriving the regex from the grammar's vocabulary with an empty-vocabulary guard.
- `schemas.ts` — `registerEventType` delegates; the hand-written regex is gone.
- `event-grammar-concessions.ts` — `word-separator:_` divergence **retired, not re-dated**.
- `docs/migrations/2026-08-10-event-name-grammar.md` — what breaks, what starts working, what an affected user does, and what happens to already-persisted streams under INV-1.
- `3f55cd0`-follow-on in the overhaul spec — see DR-27 below.

## Test Plan

**Orchestrator-run verification on the merged branch, independent of the implementer's report:**

- **Baseline** — the four affected test files: **407 pass / 0 fail**.
- **`STALE_SEED_ENTRY` tooth** — I re-asserted `divergesFromShippedPattern: true` against the repaired tree: **7 tests redden** and the code prints verbatim. The concession is genuinely retired and the tooth was not silenced. Its twin (`UNSEEDED_GRAMMAR_CONCESSION`) means the flip could not have been made early either.
- **Persisted replay, verified at source.** This is the claim that mattered most, so I checked it against the real stores rather than the report: querying both live SQLite databases **read-only**, the union of distinct event names is **79** — and the committed corpus diffs **byte-identical** against that union. It is the entire live population, not a curated selection. The grammar rejects **0** of them.
  - Row counts drifted from the implementer's 12,383 / 507 to 12,396 / 503, because *this orchestration session has itself been appending events to the main store*. Distinct-name counts match exactly (73 and 39), which is the quantity the corpus pins.
  - One persisted name (`init.executed`) is not in the live catalog, which is itself evidence the corpus is an independent read rather than a restatement of `getValidEventTypes()`.
- **Kill fixture both ways** — reverting `registerEventType` to the retired regex reddens 7: a name the old pattern admitted now throws *with the migration named*, and a snake_case name the old pattern refused now registers.
- **Suites** — root **1660 pass / 1 fail**; the single failure is `BuildBinary_EmbedsSourceState_AgreeingWithIndependentGitVerdict`, flagged independently by **two** agents in this epic as racing `npm run build`, and it passes **10/10** in isolation. MCP **26 failures — the identical known-red 9-file cluster** as Waves 3 and 4, `store.race.test.ts` included (it sits inside this task's blast radius and matches base member-for-member). Typecheck clean in both packages; no generated-artifact drift.
- **All 54 `grep-gates` steps** run locally and pass.

### DR-27, and a call the implementer correctly refused to make

075's repair drove the overhaul spec's `event-name-pattern-divergence` premise from 25 to 0, so `check-measured-premises` — blocking in CI — went to FAIL.

The implementer **declined to zero the literal in place**, arguing that the sentence around it explains a defect and rewriting it to "they disagree on 0 of 171 live names" would describe that defect while reading as though it had never existed. That was the right call, and it surfaced the decision instead of taking it.

Resolved here by keeping the **25 as a dated measurement** — the discipline DR-8 asks for elsewhere in this programme — and adding a closure note carrying the live derived `0` inside the marker the gate reads. The spec now records both what was true on 2026-08-06 and what is true now.

### Known gaps, stated rather than smoothed over

- **The no-digits clause was adopted, not merely inherited.** Zero measured counterexamples across three populations: 171 registered names, 79 persisted names / ~12.9k rows, and every user-config event name this repo registers or documents. The cost is named in the migration note rather than assumed away, and `workflow.started2` stays in the malformed fixture table so any future widening is a visible edit.
- **Reach of the measurement.** Nobody can measure event names in *external* user configs from here. The migration note therefore states the break precisely rather than claiming no one is affected.
- **Not verified:** Windows behaviour (no local lane); `npm run validate`, mutation and coverage gates were not run.
- The implementer touched two files beyond its declared set — `event-grammar-census.ts` (comment-only; three prose blocks its change made false) and that census's own test (six cases that measure the divergence and cannot stand once it is zero). Both reported rather than hidden.

https://claude.ai/code/session_01XKWiFqQUaNH4tSPrRbdh4Q


rsalus:
## Shepherd pass 2 — #1809

Zero unresolved inline threads. Two of the three "🧹 Nitpick" items from the review summary — never posted as threads, so never counted — are fixed in `32cc2c4`:

- **`event-name.ts` 601-613** — real, and the sharpest item in the stack's nitpick inventory. `buildEventNamePattern` validated ordering and the floor but not integrality, and JavaScript does not read `{1.5,2}` or `{1,Infinity}` as a quantifier: the braces degrade to literal characters, so the pattern stops constraining segment count and starts demanding that text appear in the name. Exactly the failure `EmptyGrammarVocabularyError` guards against one line above — a validator that quietly stopped validating. `Number.isInteger` rejects fractions, Infinity and NaN in one predicate. Kill-probed: disabling the guard reds the new arm. Production call sites pass the module's own constants, so this was latent, not live.
- **`event-name.test.ts` 320-326** — `dotSeparated` renamed to `plusSeparated`. The separator under test is `+`, in the one test whose whole subject is which character gets escaped.

The third (`schemas.test.ts` 2125-2154 — the retired event-name regex transcribed four times across three test files) is inventoried in #1811; it is a cross-file extraction rather than a local edit.

rsalus:
@coderabbitai approve

Issue #1838, opened 2026-08-19
bug: roots-based featureId inference injects into schemas that forbid it — exarchos_event writes are unusable over MCP


rsalus:
## Summary

Roots-based `featureId` inference (#1290) injects `featureId` into the args of **every** action not on a 3-item skip list. For actions whose Zod schema does not declare `featureId`, the strict-schema check then rejects the call the server itself mutated. The result: **`exarchos_event` is unusable from any MCP client that declares the `roots` capability, except `action: "describe"`.**

This blocks the entire delegate-phase event contract, which is mandatory: `task.assigned`, `team.spawned`, `team.task.planned`, `team.teammate.dispatched` all have to be emitted via `exarchos_event`, and `prepare_delegation` counts `task.assigned` events to populate `taskCount`.

## Reproduction

From Claude Code (MCP client that declares roots), with any active workflow:

```jsonc
// exarchos_event
{ "action": "append",
  "stream": "session-history-drill-card",
  "event": { "type": "task.assigned", "data": { "taskId": "T-01", "title": "…" } } }
```

Actual:

```
INVALID_INPUT: exarchos_event/append: unrecognized parameter(s): "featureId"
(declared by no action on exarchos_event). append accepts: event,
expectedSequence, idempotencyKey, stream. Remove the parameter or dispatch the
action that declares it — it would otherwise be ignored while the call reported
success.
```

`featureId` **is not in the caller's payload.** Same failure for `batch_append` and `query`. `action: "describe"` succeeds.

Expected: the event is appended.

## Root cause

`src/dispatch/core/dispatch.ts` ~L773–792:

```ts
if (
  rest.featureId === undefined
  && ctx.capabilityResolver !== undefined
  && ctx.rootsClient !== undefined
  && !NO_WORKSPACE_RESOLUTION_ACTIONS.has(actionName)
) {
  …
  if (resolution.success) {
    rest = { ...rest, featureId: resolution.featureId };   // ← injected here
  }
```

with, ~L367:

```ts
const NO_WORKSPACE_RESOLUTION_ACTIONS: ReadonlySet<string> = new Set([
  'describe', 'runbook', 'agent_spec',
]);
```

Two problems:

[...]

PR #1843, opened 2026-08-20
refactor(dispatch): one table and one gate for every inferred value

rsalus:
# One table and one gate for every inferred value

**Stacked on #1842** — based on `worktree-preview4-bugfix-cluster`, so this diff is the refactor
alone. Merge #1842 first. If it squash-merges, this needs rebuilding as main + this delta.

## Summary

#1842 made dispatch consult the receiving action's schema before merging an inferred `featureId`
into the payload. That closed the **instance**. It left the **class** open: the gate was hand-wired
for one field, so a second inference could merge into the payload itself and never consult a
schema — which is exactly how the fault arose the first time.

Inferred values now live in one table, and every entry passes through one gated path. Adding an
inference is a table entry, so it inherits the gate instead of re-deriving it.

## Exploration corrected the plan

Two premises of the original outline turned out to be wrong, and both are worth stating because
they changed the design:

**Inferred values cannot be moved out of validation.** The outline proposed a side channel that
kept inferred values out of the validated payload. But **43 of the 59 actions that take a
`featureId` declare it REQUIRED** — inference exists precisely to satisfy that requirement. A side
channel would fail validation for exactly the callers inference serves. That constraint is what
makes the gate load-bearing rather than cosmetic: the value is merged into a strict schema's
input, so merging it into a schema that does not declare it manufactures a rejection.

**Elicitation was never part of the class.** The outline said all three inference paths move
together. Elicitation splices a field named by the action's *own* parse error, so it can only ever
supply something the action declares. It is structurally safe already and is untouched here.

## Changes

`src/dispatch/core/inferred-values.ts` (new) owns:

- `InferenceOutcome` — `resolved` | `ambiguous` | `unavailable`. Ambiguity is a first-class outcome
  because the caller can act on it; `unavailable` covers both "nothing matched" and "the resolver
  failed", which are the same to dispatch (fall through to ordinary validation).
- `INFERRABLE_FIELDS` — the table. One entry today: `featureId` via `resolveWorkspace`.
- `applyInferredValues()` — the shared path. Three skips, applied per entry rather than per field:
  an explicit caller value always wins; an action whose schema omits the field is left alone (**the
  gate**); the latency list short-circuits pure introspection.

`dispatch.ts` calls it once and keeps no per-field logic. `actionAcceptsInferredFeatureId`
generalizes to `actionAcceptsInferredValue(action, field)`.

## Behavior

Unchanged by design. Same gate, same latency skips, same multi-match envelope with its
`validTargets` and correlation `_meta`, same best-effort fallback when the resolver fails.

## Test Plan

**Oracle integrity.** The #1838 guards are the characterization oracle. `git diff HEAD~1 -- tests/`
shows every change in that file is a predicate rename or import move — **no assertion value,
matcher, or message string was altered, and nothing was deleted**. `LATENCY_SKIP` is now derived
from the shipped table instead of being a hardcoded copy, which strengthens the oracle.

**Class closure, proved with a synthetic field.** The new tests drive `applyInferredValues` with a
*second, invented* inferrable field, so what they pin is "any table entry inherits the gate" rather
than "featureId happens to be gated" — including that an action omitting the field never receives
it, an explicit caller value is never overwritten, the skip list short-circuits before the resolver
runs, and the ambiguity envelope survives.

**Kill probe.** Removing the gate from the shared path turns the end-to-end guard red naming 21
actions.

| | Failed | Total |
|---|---|---|
| Baseline `main` | 37 | 14103 |
| This branch | **35** | 14139 |

The one delta against main was the reference-census drift tolerance (`<50` tracked files, at 51
cumulative across both branches); regenerated like the other oracles. The remaining
worktree-inventory failure is environment-dependent and fails identically on `main`.

Typecheck green. Layer boundaries green — the new module sits in the same `dispatch` layer, so no
new layer edge. Knip reports no unused exports in it. The DR-14 cast ratchet is flat: non-null
assertions were avoided by having the resolver re-check its own precondition instead of trusting
that `isAvailable` ran — which also removes a dependency one refactor away from being false.

## Docs

No prose documents this behavior (`CLAUDE.md` does not mention roots or inference, and no
`content/` or `docs/` file covers it), so nothing to update. The constraint is documented where it
lives — the module header states why the value must reach validation, why the gate is load-bearing,
and why elicitation is exempt. A prose copy would be the duplication `CONTEXT-AUTHORING` warns about.

https://claude.ai/code/session_018RKfa5AQaR8Ay3XwjQMfJr

### DIFF
--- a/src/dispatch/core/dispatch.ts
+++ b/src/dispatch/core/dispatch.ts
@@ -51,7 +51,7 @@ import {
   findIgnoredParameters,
   buildIgnoredParameterError,
 } from '../undeclared-parameters.js';
-import type { ToolAction } from '../../registry.js';
+import { applyInferredValues } from './inferred-values.js';
 import path from 'node:path';
 import {
   detectActiveStoreDivergence,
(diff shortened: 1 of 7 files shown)


rsalus:
bugbot run

rsalus:
@coderabbitai review

rsalus:
@coderabbitai approve

rsalus:
@coderabbitai approve

Issue #1822, opened 2026-08-17
Epic: effect ledger & emission — anchors 028, 036, 039 (root of the deferred graph)


rsalus:
## Summary

Spun out of #1763 on 2026-08-17, when the authority-topology census was re-measured against the landed tree. This epic owns the **effect ledger and the emission contract** — the root of the deferred anchor graph.

**Measured status (`main` @ `2919547d5`):** `EffectPlan` had **no `emits` field**, and that module referenced the event store **zero times**. Nothing had started.

**[corrected — re-baselined 2026-08-20, `main` @ `e63493269`] Work HAS started and one member is closed.** #1832 landed `emits` on `EffectPlan`, armed enforcement (`events.emission-enforcement` defaults to `block`), and delivered the `event` reachability hop — **closing #1826**. Anchor 028 (#1765) is now **partial**: the declaration half is bound, the **sink half is not** — `effect-carrier.ts` still references the event store zero times.

## Plan (2026-08-20)

Spec: `lvlup-sw/docs:exarchos/docs/specs/2026-08-20-effect-ledger-closure.md` — workflow `effect-ledger-closure`, **7 design requirements, 22 tasks**, all four plan gates green (coverage 7/7, provenance 7/7 with 0 orphan refs, decomposition 22/22 with a valid DAG and no file conflicts, spec-coverage 10/10). Currently at `plan-review`.

**The finding that shaped it: the emission contract is optional at every layer, so all three shipped mechanisms can report clean over a population that opted out.**

| Layer | Field | Consequence |
|---|---|---|
| Carrier | `readonly emits?:` on `EffectPlan` | *"a plan that records nothing stays a legal plan"* |
| Contract | emission field on `ContractDeclaration` | *"existing declarations predate this field and declare none, so the emission axis has nothing to verify"* |
| Registry | `autoEmits` on an action | **68 of 124 actions declare none** |

Undeclared by tool: `exarchos_orchestrate` **34**, `exarchos_view` **26 of 26**, `exarchos_event` 4, `exarchos_workflow` 3, `exarchos_sync` 1. The view actions are read-only, so emits-nothing is very likely correct *in substance* for all 26 — it is simply never *stated*, which is what makes abstention indistinguishable from judgment.

[...]

Issue #1810, opened 2026-08-10
gate-runner: same-operation recovery rebuilds the skip descriptor from the retry, not the persisted evidence


rsalus:
## Summary

On a same-operation retry, `runPhaseGate` rebuilds the `gate.executed` signal from the **retry's** provider result rather than from the evidence row the first attempt persisted. If the two disagree about skip metadata, the signal describes the persisted `indeterminate` evidence incorrectly.

Raised by CodeRabbit on #1804 as an *outside diff range* comment (🗄️ Data Integrity, 🟠 Major). It could not be posted as an inline thread due to platform limits, so it never appeared in the PR's thread list and was not addressed in that stack.

## The defect

`servers/exarchos-mcp/src/orchestrate/gate-runner.ts`, the `sameOperation !== undefined` branch:

```ts
await appendGateExecutedSignal(
  dependencies.eventStore,
  request.streamId,
  operationId,
  provider,
  sameOperation.record,          // <- evidence from the FIRST attempt
  readGateSkipDescriptor(providerResult),   // <- skip data from THIS attempt
);
```

`sameOperation.record` is the durable evidence from the earlier attempt; `providerResult` is the current one. The skip descriptor (`skipped`, `discriminant`, `reason`) is spread into `data.details` of the emitted `gate.executed`, so the row can carry skip metadata that never belonged to the evidence it references.

## Reachability

Narrower than it first looks, and worth stating precisely so this is scoped honestly:

- The append is idempotent on `gate.executed:${evidence.evidenceId}`. If the first attempt's signal landed, the retry collapses onto that row and nothing diverges.
- The exposed window is: **evidence append succeeded, signal append failed**, then a retry under the same `operationId` whose provider returns a *different* skip descriptor (skip reason changed, or it no longer skips at all).

So it needs a partial failure plus provider non-determinism across attempts. Not hypothetical — that combination is exactly what the same-operation recovery path exists to serve — but not an everyday path either.

Notably this is the same failure shape DR-7 already legislates against inside one attempt: the module docstring says the row's `passed:false` and its `details.skipped:true` "cannot tell different stories". Across a retry boundary, they currently can.

## Why it wasn't fixed in the #1764 stack

[...]

Issue #1774, opened 2026-08-09
IMO anchor 037: Derive PHASE_EXPECTED_EVENTS


rsalus:
> **Anchor task — not yet decomposed.** Per the spec: *"Each anchor carries its DR's provenance and a re-plan trigger. They are not implementation tasks; they exist so provenance resolves and so the re-plan pass has an explicit entry point. Decomposing these before DR-6's census reports its real subject list would fabricate rather than derive the work."* Decompose via a `/plan` pass against the census output, not by hand.

## Task 037 (as specified)

**Risk Tier:** high · **Boundary Touching:** true · **Implements:** DR-16 · **Dependencies:** 031 · **Parallelizable:** No
**Detail:** Delete `PHASE_EXPECTED_EVENTS` as a hand-maintained artifact and derive it from the union of `autoEmits` across each phase's reachable actions plus T4 workflow declarations. No built-in phase name may appear as a literal key in substrate code (INV-6). `_eventHints.missing` is recomputed from the derived set, with a golden fixture pinning current output.
**Re-plan input:** G5 census output — which phases currently key off literal built-in names.

### DR-16: Derive `PHASE_EXPECTED_EVENTS` **[T-9]**

**Acceptance criteria:**
- The table is **deleted as a hand-maintained artifact**, derived from the union of `autoEmits` across the phase's reachable actions plus T4 declarations.
- **No built-in phase name appears as a literal key in substrate code** (INV-6).
- `_eventHints.missing` is computed from the derived set; a golden fixture pins current output so behavior is unchanged for existing phases.

---

Source: `docs/specs/2026-08-06-internal-mechanics-overhaul.md` task 037.
Epic: https://github.com/lvlup-sw/exarchos/issues/1763

**Re-plan precondition:** the authority-topology table must be re-measured against the landed tree first — see the topology-staleness issue. The current census reports rows that PR #1755 already closed, so planning against it would fabricate completed work.

---

## Re-plan disposition (2026-08-17)

Epic #1763's stated precondition — re-measure the authority-topology census against the landed tree — was executed against `main` @ `2919547d5`. Scope was rewritten and the anchors re-parented. Source: `lvlup-sw/docs:exarchos/docs/specs/2026-08-17-imo-anchor-rescope.md`.

**Disposition: RETAINED in #1763.**

**Measured 2026-08-17:** `PHASE_EXPECTED_EVENTS` is a hand-maintained literal at `src/verbs/gates/check-event-emissions.ts`.

[...]

Issue #1768, opened 2026-08-09
IMO anchor 031: Contract meta-model tightening (isolated PR)


rsalus:
> **Anchor task — not yet decomposed.** Per the spec: *"Each anchor carries its DR's provenance and a re-plan trigger. They are not implementation tasks; they exist so provenance resolves and so the re-plan pass has an explicit entry point. Decomposing these before DR-6's census reports its real subject list would fabricate rather than derive the work."* Decompose via a `/plan` pass against the census output, not by hand.

## Task 031 (as specified)

**Risk Tier:** high · **Boundary Touching:** true · **Implements:** DR-10 · **Dependencies:** 013 · **Parallelizable:** Yes
**Note:** Lands as its own PR — `contract/` is under active change.

### DR-10: Contract meta-model tightening **[T-10]**

**Acceptance criteria:**
- `AutoEmitSpecSchema.event` changes from `z.string()` to a catalog-validated `EventTypeRef`; `tier` and `coupling` are added to `EvidencePolicy`. **[path corrected 2026-08-20]** The module is `src/contract/compiler/meta-model.ts` — #1817 added the `/compiler/` segment, and `src/contract/meta-model.ts` has not existed since. The premise itself is re-confirmed against `main` @ `e63493269`: `event: z.string()` still sits at line 93 of the moved file, and `EvidencePolicy` lives in that same module rather than a separate `evidence-policy.ts`.
- A stale `autoEmits` row naming an unregistered type fails compilation.
- Compilation remains byte-stable across repeated runs (P03-03).
- **Lands as its own PR**, separate from the waves — `contract/` is under active change.

**Wave 3 — Observation**

---

Source: `docs/specs/2026-08-06-internal-mechanics-overhaul.md` task 031.
Epic: https://github.com/lvlup-sw/exarchos/issues/1763

**Re-plan precondition:** the authority-topology table must be re-measured against the landed tree first — see the topology-staleness issue. The current census reports rows that PR #1755 already closed, so planning against it would fabricate completed work.

---

## Re-plan disposition (2026-08-17)

Epic #1763's stated precondition — re-measure the authority-topology census against the landed tree — was executed against `main` @ `2919547d5`. Scope was rewritten and the anchors re-parented. Source: `lvlup-sw/docs:exarchos/docs/specs/2026-08-17-imo-anchor-rescope.md`.

**Disposition: RETAINED in #1763.** Not spun out, and deliberately so.

[...]

PR #1849, opened 2026-08-20
docs: the phase-gate taxonomy, measured from the live tree

rsalus:
## Summary

Adds a standalone architecture study of the phase-gate surface, measured from the live tree rather than transcribed from prose, plus the two contract entries that make it tracked and keep the suite green. Every count in the document was extracted by importing the live modules and enumerating them — `TOOL_REGISTRY`, `EVENT_EMISSION_REGISTRY`, the five `create*HSM` factories, the four policy modules, and the effect ledger's own `detectModuleEffects`. It exists because "which gates run, what they record, and what enforces them" currently has four different answers depending on which declaration surface you ask.

Sections 1–7 are measurements. Sections 8–10 are an explicitly-labelled proposal: a target state, eight refactors, and how they compose with epic #1763's re-scoped anchor graph.

## Changes

- **docs/phase-gate-taxonomy.html** — the study. Self-contained: no external requests, both diagrams hand-authored inline SVG, light/dark aware, every chart carries a table view.
- **tools/audit/prose-manifest.ts** — registers the document in `RETAINED`. **This is a policy decision for review, not a formality** (see below).
- **tests/architecture/top-level-contract.test.ts** — declares `.lavish` in `ALLOWED_DOT_DIRS`, fixing a pre-existing local-only red.

## Test Plan

Ran the architecture tier, which owns every contract this touches. `prose-exodus` (13/13) and `top-level-contract` (19/19 with `prose-exodus`) pass. The document itself was rendered in headless Chromium in both colour schemes and audited programmatically for console errors, horizontal overflow and clipped content at 380 / 768 / 1280px — all clean.

---

### What the study establishes, on merged main @ `e63493269`

| | |
|---|---|
| 29 live atomic phases across 5 HSMs | → 6 phase kinds → 4 gate resolvers |
| `MERGE` and `GATHER` carry `gates: null` | 9 of 29 phases ungated by construction |
| 28 of 124 registered actions carry gate metadata | 14 blocking, 14 advisory, 9 with a `gateClass` |
| Three execution tracks | 5 mint `gate.executed` from persisted evidence · 17 self-emit · 1 emits a non-gate event · **5 record nothing** |
| Of those five silent gates | **two are blocking** — `validate_pr_stack`, `debug_review_gate` |
| 8 of 18 resolved obligations | leave no record of their own |

Four findings worth separate attention:

1. **The four resolver families speak four vocabularies.** Only `ladder` and `plan` name registry actions; `review` names skill folder names and `synthesis` names leg labels that exist only inside `prepare_synthesis`. Any cross-family coverage metric or admission predicate has to special-case all four.
2. **`generate_traceability` is a gate to one authority and not the other** — a member of the plan-structure sequence at the standard and deep rungs, with no gate metadata and no emission.
3. **The effect ledger cannot say which gate spawns a process.** All gate process effects route through `verbs/gates/gate-utils.ts`, the sole `child_process` importer among 40 gate modules and imported by 23 of them. Per-module attribution is a documented boundary; the consequence is that the ledger's answer for the whole gate surface is one file and one layer-granular owner whose idempotency contract delegates the question back to the gates.
4. **Three live atomic HSM phases are absent from `ALL_PHASES`**, one of them PLAN-kind (`overhaul-plan-review`). This is the phase-sequencing row anchor 045 already predicts — measured, with its subject list.

Measured against the invariant catalog: **7 of 8** long-running gates emit no INV-10 liveness pair, **0 of 28** emit an INV-13 intent row (including `check_test_adequacy`, which reverts the working tree), and **4 of 28** accept an INV-8 idempotency key.

### The RETAINED entry needs a decision

`tests/architecture/prose-exodus.test.ts` asserts that nothing under `docs/` is tracked unless it is retained, and the rule is explicit: a new document under `docs/` relocates by default and has to be *argued into* the list. So this PR argues it, and the argument should be accepted or rejected rather than waved through.

**For:** it is the canonical statement of the phase-gate architecture, sibling in kind to `system-design.html`, and it is read when a gate is added, retyped or re-bound. **Against:** the exodus destination is `lvlup-sw/docs`, and an architecture study is exactly the genre that moved. If the reviewer prefers the destination, the correct action is to drop both the file and the manifest entry from this PR and land it there instead — the document is self-contained and moves without edits.

### Known-failing, unrelated to this change

Two architecture tests fail in a worktree and are untouched by this PR:

- `published-contract :: EveryShippedDirectory_ResolvesAfterTheMove` — `dist/release-verify.js` is absent because no build ran in this worktree. Present in the main checkout.
- `worktree-inventory :: EveryRegisteredWorktree_IsRecorded` — 67 registered worktrees against a recorded baseline of 61. Environmental accumulation; deliberately not rebaselined here, since that would bake in unrelated stale worktrees.

---

**Results:** Architecture tier 230/232 ✓ (2 pre-existing environmental failures, above)
**Related:** #1763, #1768, #1774, #1810, #1822

https://claude.ai/code/session_013wN26gAX6YsHDzYXBK8PLm


rsalus:
@coderabbitai review

Issue #1825, opened 2026-08-17
Epic: catalog disposition & replay — anchors 041, 042 (blocked on 028 + event hop)


rsalus:
## Summary

Spun out of #1763 on 2026-08-17, when the authority-topology census was re-measured against the landed tree. This epic owns **event-catalog disposition and replay compatibility** — retiring duplicate event types and proving the log still reads.

**Why it left #1763.** Anchor 041 is gated on P07-01: zero unexplained disagreements across **at least twenty live workflows**. That is an observation criterion measured over time, not a task that can be scheduled into a bounded closure epic.

## Scope

- [ ] #1778 — anchor 041: Catalog disposition
- [ ] #1779 — anchor 042: Replay and compatibility

## Chain

```
028 → 038-event (both in the effect-ledger epic) → 041 → 042
```

**Blocked twice over** — *re-baselined 2026-08-20, and one blocker has cleared.* The `event` reachability hop **shipped in #1832** (#1826 closed), so `REACHABILITY_HOPS` carries it today. What still blocks 041 is anchor 028's effect ledger, which is **partial**: `emits` is bound on `EffectPlan` but `effect-carrier.ts` references the event store zero times, so the sink half is unbound. One blocker remains, not two. **[re-measured 2026-08-20, second pass]** The "unbound sink" reasoning was wrong — `validateRegistrationWelds()` returns `ok: true` with 0 diagnostics, and both `runEffect` call sites pass a ledger recorder. 028 is still **partial**, but the remainder is its *structural teeth*: `emits` ships as **optional** where DR-7 requires it, `runEffect` returns `EffectOutcome<T>` rather than `Committed<T>`, the "fails to compile" criterion is a runtime throw, and no `role: 'primary' | 'recovery'` discriminator exists. See #1765.

## Measured status (`main` @ `2919547d5`)

- **Report-coupled event types: 25.** Unmoved since the overhaul spec was written. The Wave-5 exit target is **2** — `team.spawned` and `team.disbanded`, each carrying `blockedBy: #1473`, an owner and an ISO expiry.
- **Event-name pattern divergence: 0.** This *did* move — it was 25, and task 075 (#1809) collapsed the two authorities that decide what an event name may be. That is the one part of the event-catalog row already closed.
[...]

Issue #1823, opened 2026-08-17
Epic: reconciliation — anchors 032, 033, 034 (blocked on 028)


rsalus:
## Summary

Spun out of #1763 on 2026-08-17, when the authority-topology census was re-measured against the landed tree. This epic owns **reconciliation**: observing real-world state, recording divergence, and deciding which authority wins.

**Why it left #1763.** Reconciliation is an unbuilt subsystem, not a bounded closure. The re-plan's admission rule was that #1763 keeps only work whose subject list the census can enumerate today. Reconciliation's cannot be enumerated — there is no reconciler interface to measure, so any task breakdown would fabricate rather than derive. That is the same failure mode the original epic warned about when it deferred these anchors in the first place.

## Scope

- [ ] #1769 — anchor 032: Reconciler interface + content-addressed observation
- [ ] #1770 — anchor 033: Boundary-triggered reconciliation (no daemon)
- [ ] #1771 — anchor 034: Divergence recording + authority precedence

## Chain

Strictly serial, and blocked at the head:

```
028 (in the effect-ledger epic) → 032 → 033 → 034
```

**None of these can start until anchor 028 (#1765) lands — and it has only half landed.** *Re-baselined 2026-08-20:* #1832 bound `emits` on `EffectPlan` and armed enforcement, but `effect-carrier.ts` still references the event store zero times. Anchor 032 needs the coupling to be observable end-to-end, so the sink half is the live blocker. Treat 028 as **partial**, not unstarted, when re-checking readiness. **[re-measured 2026-08-20, second pass]** The "unbound sink" reasoning was wrong — `validateRegistrationWelds()` returns `ok: true` with 0 diagnostics, and both `runEffect` call sites pass a ledger recorder. 028 is still **partial**, but the remainder is its *structural teeth*: `emits` ships as **optional** where DR-7 requires it, `runEffect` returns `EffectOutcome<T>` rather than `Committed<T>`, the "fails to compile" criterion is a runtime throw, and no `role: 'primary' | 'recovery'` discriminator exists. See #1765.

## Design constraint carried over, and it is load-bearing

[...]

Issue #1765, opened 2026-08-09
IMO anchor 028: Effect ledger — emits coupling on EffectPlan


rsalus:
> **Anchor task — not yet decomposed.** Per the spec: *"Each anchor carries its DR's provenance and a re-plan trigger. They are not implementation tasks; they exist so provenance resolves and so the re-plan pass has an explicit entry point. Decomposing these before DR-6's census reports its real subject list would fabricate rather than derive the work."* Decompose via a `/plan` pass against the census output, not by hand.

## Task 028 (as specified)

**Risk Tier:** high · **Boundary Touching:** true · **Implements:** DR-7 · **Dependencies:** 027 · **Parallelizable:** No
**Re-plan input:** G5 census output — which effects currently lack a coupling. First migrated consumer is `VcsMutationOwner` (the G4 kill fixture).

### DR-7: Effect ledger — emission as a precondition of the effect landing **[T-4]**

Extends the shipped `core/effect-carrier.ts` (P04-01), which carries `owner`/`idempotent`/`compensation` but **no event coupling**.

**Acceptance criteria:**
- `EffectPlan` gains a required `emits: EventType`; an effect cannot be planned without naming the event that records it.
- `runEffect` requires an appender and returns `Committed<T>`; `T` is unreachable without the append. The dry-run arm is unchanged and still appends nothing.
- A handler performing an effect without committing its event **fails to compile** (it holds an unusable carrier).
- Boot-time bijection: every `plan.emits` names a registered T1 event; every T1 event has exactly one **primary** owner (`role: 'primary' | 'recovery'`).
- Idempotency key is `<eventType>:<operationId>`, reusing `idempotency_claims` — no new storage (INV-8).
- `VcsMutationOwner` is the first migrated consumer (G4 kill fixture).

---

Source: `docs/specs/2026-08-06-internal-mechanics-overhaul.md` task 028.
Epic: https://github.com/lvlup-sw/exarchos/issues/1763

**Re-plan precondition:** the authority-topology table must be re-measured against the landed tree first — see the topology-staleness issue. The current census reports rows that PR #1755 already closed, so planning against it would fabricate completed work.

---

## Re-plan disposition (2026-08-17)

[...]

PR #1850, opened 2026-08-20
feat(events): emission as a precondition of planning an effect (anchor 028, 3/4)

rsalus:
## Summary

Finishes **three of anchor 028's four acceptance criteria** ([#1765](https://github.com/lvlup-sw/exarchos/issues/1765)).

**On the two downstream epics, stated precisely.** [#1823](https://github.com/lvlup-sw/exarchos/issues/1823) and [#1825](https://github.com/lvlup-sw/exarchos/issues/1825) both name anchor 028 as their blocker, and 028 is **not fully closed** here. #1825 enumerates four "structural teeth" it is waiting on; this PR delivers three of them (`emits` required, a `Committed<T>`-shaped return, compile-time failure). The fourth it names — *"no `role` discriminator exists"* — is the premise that turned out false. Whether three of four is enough to start 032 and 041 is for those anchors' owners to judge, and this PR does not claim it for them. The fourth — the primary-owner bijection — is **deferred on measured evidence** and recorded below as a named residual rather than reported as met.

PR #1832 built the effect↔event coupling and it works. What was missing was the *obligation*: `emits` was optional, so the guarantee held by the good behaviour of two call sites rather than by construction, and the third call site nobody has written yet was free to skip it. This closes that.

- **`EffectPlan.emits` is required**, as a tagged union — `records(...)` (non-empty by type) or `recordsNothing(because)`. "Records nothing" and "nobody decided" are now different values instead of the same silence.
- **The committed value carries its evidence.** The success arm cannot be built without something saying the record exists, so obtaining `T` means the append happened.
- **The recorder is demanded of every live run** and its parameter narrows to a required `EmissionRecorder`, so omission is a compile error rather than a runtime throw. The throws stay as defence in depth.

## Changes

| Area | What |
|---|---|
| `dispatch/core/effect-carrier.ts` | Required `emits`; `records`/`recordsNothing`/`declaredEmissions`; two-armed branded `EmissionEvidence`; recorder narrowed; 6 new `@proof` aliases |
| `vcs/mutation-owner.ts` | Ledger set migrated; the idempotency-replay arm carries a **witness** instead of a fabricated receipt |
| `install/atomic-promotion.ts` | Promotion set migrated; recorder required; live-mode guard clause |
| `tools/conformance`, `tools/audit` | Two live authority modules re-trued — both asserted `emits` was optional |
| `tests/acceptance/*` (new ×2) | Spawned-`tsc` compile gate; kill-probe suite |
| `tests/unit/dispatch/core/effect-carrier.test.ts` | DR-5 edge tests; two shipped tests inverted |

### Three things worth review attention

**The replay path had no defined semantics, and it is the interesting part.** `mutation-owner.ts` returns a committed value for a terminal recorded on a *previous* run, where no receipt can be minted. Forcing it to fabricate one would claim this run wrote the fact. It carries a witness read from the ledger fold instead — and that witness is **branded by the carrier**, because an unbranded one would let any object literal buy a committed value, reopening the exact hole the branded port closed. The trust model is stated in the module rather than implied — including the part that is NOT parity: the brand stops an object literal from becoming evidence, but `emissionRecorder` is strictly stronger, because it forces a sink to be invoked and awaited while the witness constructor takes two strings and consults nothing. That asymmetry is the price of the replay path, and it is written down rather than glossed.

**`promoteTree` needed a guard clause.** The owner wraps its caller's recorder in a capability, so a wrapper around nothing satisfies the carrier's brand check and the refusal would have landed at the *success terminal* — after the tree had already been promoted. Scoped to live mode: a dry-run promotes nothing and records nothing, so demanding the capability there would break the dry-run guarantee.

**The compile claims live in `src/`, not in a test.** `tests/unit/**` is read by no `tsc` program — the root config excludes `**/*.test.ts` and `tests/tsconfig.json` excludes `unit/**`. Measured both ways: a deliberate type error in `tests/acceptance/` reddens `npm run typecheck`; the identical error in `tests/unit/` passes it clean. Every compile assertion is therefore an exported `@proof` alias in the source module.

## Anchor 028 — criteria re-run and measured

| Criterion | Result |
|---|---|
| `EffectPlan` gains a **required** `emits` | ✅ `readonly emits: PlanEmissions` (`effect-carrier.ts:214`) |
| `T` unreachable without the append | ✅ success arm carries required `evidence` |
| Effecting without committing **fails to compile** | ✅ 13 `@proof` aliases + a spawned-`tsc` fixture |
| Every T1 event has one **primary** owner | ⛔ **DEFERRED — named residual** |

### Why the fourth is deferred, and who owns it

The requirement was authored on a false premise: it claimed `role: 'primary' \| 'recovery'` "appears nowhere in the tree". It does — `AutoEmissionRole` (`registry/gate-metadata.ts:24`), with **86 live edges** carrying both `role` and `owner` across 54 events. The gap was never the discriminator; it was the bijection, and the bijection has two possible populations and both fail:

- **Over the plan-declared events** it is **vacuous**. Zero `autoEmits` anywhere name `vcs.*` or `promotion.executed`, so this slice would author all four triples itself and then have the producers derive from that same table. A gate whose input its own slice authors cannot fail by drift.
- **Over the live 86 edges** it is **red on arrival and probably mis-keyed**: 1 event with zero primaries (`merge.recovered`) and 4 with more than one — `gate.executed` carrying **24**. That 24 looks correct, not broken: many gates each emit it, each the primary emitter of its own occurrence. The criterion keys on event *type*; the catalog keys on event *occurrence*.

Deciding that key is a modelling question about the real catalog. **Owner: the emission-denominator slice.**

## Test Plan

Every gate added here was **killed on purpose and observed to fail** before being kept — the property this programme keeps failing to establish.

- `npm run typecheck` — clean across all three programs (root, `tools/conformance`, `tests`)
- `npm run build` — clean; no git drift, so `render:guard`'s render half is satisfied
- `npx vitest run tests/unit/{dispatch,vcs,install}` + `events/schemas` — **1708 passed**
- `npx vitest run tests/unit/events tests/unit/verbs/team` — passed
- `npm run test:conformance` — authority suites pass (28)
- `npx vitest run tests/architecture` — 230/232 (see below)
- New: `tests/acceptance/effect-carrier-compile-gate.test.ts` (3), `tests/acceptance/effect-plan-teeth-kill-probes.test.ts` (6)

**Kill probes, each asserting both sides** — the real carrier rejects the bad program, and a relaxed copy accepts it:

| Relaxation | Observed |
|---|---|
| `emits` back to optional | the omission compiles |
| `evidence` optional on success arm | an evidence-free value compiles |
| brand removed from the witness | a hand-built witness compiles |
| recorder back to `\| undefined` | a capability-free live call compiles |
| `declaredCount > 0` condition restored | **a capability-free live run COMMITS** (executing probe, spawned node) |
| truncation guard removed | 3 unit tests redden |
| dry-run early return removed | 6 unit tests redden |

No probe touches the live tree; all relax text in a temp directory, and a case asserts the source still declares every guard afterwards.

### Known-red, verified not caused by this change

- `tests/unit/events/store.race.test.ts` — fails **identically with baseline sources reverted** (checked out `e63493269` for the three changed files and re-ran). Known local merge-orchestrate redness.
- `tests/architecture/worktree-inventory.test.ts` — counts live worktrees on the machine; drifts with concurrent sessions.
- `render:guard` — fails in a git worktree only, looking for `node_modules/tsx/dist/cli.mjs` under the worktree root while resolution walks up to the parent checkout. Unrelated to this diff.

## Adversarial review — failed once, remediated

A fresh-context reviewer failed the first pass on **1 HIGH + 3 MEDIUM + 2 LOW**. All six are fixed in `75b48ba60`; the HIGH is worth reading.

**The recorder gate had a runtime half that no probe measured.** DR-5's probe for it was compile-only, and the probe file's header claimed all four gates were type-level. That is false for this one: the parameter is the type-level half, but `effect-carrier.ts`'s brand check is the runtime half — its own comment calls itself *"the boundary the type system does not govern"*. Restoring `declaredEmissions(plan).length > 0 &&` keeps the parameter required, so the compile probe and the type proof both stay green **while the abstention hole reopens for every `records-nothing` plan**. There is now an **executing** probe: it emits the relaxed copy to JavaScript, runs it in a spawned node process, and reads the outcome off stdout. Real carrier → `refused`; relaxed copy → `committed`.

Writing that probe exposed a defect in the probe harness itself. The capability check is spelled **identically** in `recordEmissions` and in `runEffect`, and `String.replace` takes the first match — so the probe relaxed a guard it was not asserting about, and reported that the gate held. `materialize()` checked presence but not **uniqueness**; it now rejects an ambiguous target. That is the same vacuity class the file exists to catch, caught by the file on itself.

The rest: the probes' control arms asserted only that compilation failed and not that it failed *for the fixture* (the arms are asymmetric, so an unrelated error would have made all four pass while measuring nothing); `schemas.test.ts` still called `promoteTree` with two arguments after the signature narrowed to four, passing only because the dry-run arm returns early; the replay witness's brand was probed but its **wiring** was not; and the module overclaimed the two evidence arms as equally strong — `replayedEvidence` takes two strings and consults nothing, while `emissionRecorder` forces a sink to be invoked and awaited. The asymmetry and its reason are now written down rather than papered over.

https://claude.ai/code/session_0191kQ7YQZ8khTpiGUmDja2y

### DIFF
--- a/src/dispatch/core/effect-carrier.ts
+++ b/src/dispatch/core/effect-carrier.ts
@@ -115,6 +115,66 @@ export interface EffectEmission {
 }
 
 /**
+ * A plan that genuinely records no fact, and the reason it does not.
+ *
+ * This arm exists so that "records nothing" and "nobody decided" are different
+ * values rather than the same absence. An optional field could not tell them
+ * apart: a missing `emits` meant both, which is what let a plan promise no
+ * record by saying nothing at all.
+ *
+ * `because` is required for the same reason the arm is: an abstention with no
+ * stated reason is indistinguishable from an oversight, and the whole point of
+ * naming the abstention is that somebody had to decide it.
+ */
+export interface RecordsNothing {
+  readonly kind: 'records-nothing';
+  readonly because: string;
+}
+
+/**
+ * A plan that records, carrying the emissions it promises.
+ *
+ * The tuple type is non-empty by construction, so `records()` with nothing in
+ * it does not compile. An empty emission list is an abstention wearing the
+ * declaring arm's clothes, and it would reintroduce exactly the ambiguity
+ * {@link RecordsNothing} exists to remove.
+ */
+export interface RecordsEmissions {
+  readonly kind: 'records';
+  readonly emissions: readonly [EffectEmission, ...EffectEmission[]];
+}
+
+/**
+ * What a plan promises to record. Required on every {@link EffectPlan}: an
+ * effect that cannot say what records it is a state mutation that is not an
+ * event, which is the hole INV-1 names.
+ */
+export type PlanEmissions = RecordsEmissions | RecordsNothing;
+
+/** Declare the emissions a plan records. At least one, by type. */
+export function records(
+  first: EffectEmission,
+  ...rest: readonly EffectEmission[]
+): RecordsEmissions {
+  return { kind: 'records', emissions: [first, ...rest] };
+}
+
+/** Declare that a plan records nothing, and why. */
+export function recordsNothing(because: string): RecordsNothing {
+  return { kind: 'records-nothing', because };
+}
+
+/**
+ * The emissions a plan declares, flattened across both arms.
+ *
+ * The one accessor every consumer reads, so the arm distinction stays inside
+ * this module rather than spreading a `kind` check through every caller.
+ */
+export function declaredEmissions(plan: EffectPlan): readonly EffectEmission[] {
+  return plan.emits.kind === 'records' ? plan.emits.emissions : [];
+}
+
+/**
  * The typed plan for a single effect. It encodes the three things the plan
  * mandates for every effect:
  *   - `owner`        — the single typed owner accountable for the effect;
@@ -141,12 +201,17 @@ export interface EffectPlan {
    * to tell "the intent landed" from "a terminal landed" — the exact
    * distinction that says whether an interrupted run needs convergence.
    *
-   * Optional, so a plan that records nothing stays a legal plan and every
-   * construction site that predates the emission axis still compiles. An
-   * undeclared `emits` means "this plan promises no record", never "the record
-   * is inferred from `effectClass` or `owner`"; see the vocabulary note above.
+   * REQUIRED. A plan cannot be built without saying what records it, because a
+   * plan that records nothing by saying nothing is the INV-1 hole: an effect
+   * that lands without its event is a state mutation that is not an event.
+   *
+   * An effect that genuinely records nothing is still expressible — it declares
+   * {@link recordsNothing} and carries its reason. What is no longer
+   * expressible is the silence that used to mean both that and "nobody
+   * decided". Nothing here is inferred from `effectClass` or `owner`; see the
+   * vocabulary note above.
    */
-  readonly emits?: readonly EffectEmission[];
+  readonly emits: PlanEmissions;
 }
 
 /**
@@ -160,7 +225,7 @@ export function emissionsWhen(
   plan: EffectPlan,
   when: EmissionCondition,
 ): readonly EffectEmission[] {
-  return (plan.emits ?? []).filter((emission) => emission.when === when);
+  return declaredEmissions(plan).filter((emission) => emission.when === when);
 }
 
 /**
@@ -387,10 +452,12 @@ export class UnrecordedEmissionError extends Error {
  * Record every emission a plan declares for one condition, in order, and return
  * the receipts.
  *
- * A plan declaring nothing for the condition records nothing and needs no
- * capability — that is what keeps a plan with no `emits` inert. Once a plan does
- * declare, though, the run may only continue on evidence: one minted receipt per
- * declared emission. Anything less throws.
+ * A plan declaring nothing for THIS condition records nothing for it — which is
+ * ordinary: an intent-only plan reaches its terminals with nothing to append.
+ * That is a statement about one condition, not about the plan: every plan now
+ * declares, and a plan that records nothing at all says so in its own arm.
+ * Once a condition does declare, the run may only continue on evidence: one
+ * minted receipt per declared emission. Anything less throws.
  */
 async function recordEmissions(
   plan: EffectPlan,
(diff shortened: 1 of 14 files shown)


rsalus:
@coderabbitai All three addressed in `407b877f5` (and one was already fixed in `75b48ba60`, which landed after the review snapshot).

**1. Planning ordinals in comments — fixed, and it was a straight convention violation on my part.** Every ordinal this branch introduced is gone. The comments now state the constraint instead of citing the identifier for it — e.g. the emission field's doc says *"a state mutation that is not an event — a change to the world that the event log does not know happened"* rather than naming the invariant. The proof-block header is renamed to an ordinal-free string and both probe files that key on that text moved with it, which was the part worth being careful about.

Pre-existing ordinals in `atomic-promotion.ts` and `mutation-owner.ts` are left alone — they predate this branch and are not mine to rewrite. Your finding scoped to the three files I authored, which is correct.

**2. `worktree-provisioner.test.ts` missing `emits` — fixed, and thank you, this was a real miss.** The migration task named that file explicitly and I skipped it. It passed anyway because `tests/tsconfig.json` excludes `unit/**` and `plannedDryRun` just hands the plan back — which is exactly the blind spot this PR is about. The literal now declares `records({ event: 'vcs.executed', when: 'on-success' })` and is typed `EffectPlan` so the annotation is load-bearing rather than inferred.

**3. Truncation marker guard — already fixed** in `75b48ba60`, before this review posted. `materializeCarrier` now throws a named error when the marker is absent instead of letting `slice(0, -1)` silently keep both proof blocks. The sibling probe file gained the same guard plus a **uniqueness** check, after the executing probe caught `String.replace` relaxing the wrong site: the capability check is spelled identically in `recordEmissions` and `runEffect`, so a non-unique `find` had been editing a guard the probe was not asserting about — and reporting that the gate held.

Verification after the fixes: `npm run typecheck` clean across all three programs; 1391 tests green in `tests/unit/{dispatch,vcs,install}`; both acceptance suites (9 cases) pass, including the marker rename they depend on.

rsalus:
bugbot run

rsalus:
@coderabbitai Follow-up on the three nitpicks. Two were already fixed before your snapshot; the third was right and is done in `08abc3d6b`.

**Nitpick 1 — return the compiler output from `compiles`.** Already fixed in `75b48ba60`. It is now `compile()` returning `{ accepted, output }`, and both suites interpolate `.output` into their assertion messages. Same reasoning you gave: a probe that fails without a diagnostic says something is wrong without saying what.

**Nitpick 2 — the residue check pins an ordinal-bearing header.** The ordinal is gone as of `407b877f5`, but your deeper point stood: three sites were keying on proof-block comment text independently, so a reworded comment would break them one at a time. Now one exported constant, referenced everywhere.

**Nitpick 3 — share the compiler harness.** Correct, and this was the substantive one. `TSC_BIN`, `TSC_FLAGS`, the `execFileSync` wrapper and the materialization logic now live in `tests/helpers/carrier-compile-harness.ts`, imported by both suites. Your stated reason is the one that decided it: two copies of a flag list drift, and a drifted flag set means the suites measure different compilers — so a guard could hold under one and not the other with nothing to say so.

One design note on the shared marker, since it collapsed two values into one. The suites had been truncating at different proof blocks. The shared constant uses the **earlier** one, which removes both. That is safe in the direction that matters: the fixtures fail on the carrier's *types*, not on its proofs, so removing proofs never makes a fixture compile by itself — whereas keeping a block that asserts the relaxed property makes the copy fail for the right reason, which is indistinguishable from the guard holding.

Verification after the refactor: `tsc --noEmit -p tests` clean; both acceptance suites green (9 cases, including the executing probe); `test-inventory` and `test-tree-contract` green (22).

Separately, I corrected an overclaim in the PR body that was mine, not yours to catch: it said this PR "unblocks #1823 and #1825". Both epics name anchor 028 as their blocker and 028 is not fully closed here, so the accurate claim is three of the four teeth #1825 enumerates — the fourth being the `role`-discriminator premise that turned out false.

rsalus:
@coderabbitai full review

All findings from the first pass are addressed and the branch has moved five commits since that snapshot. Re-reviewing against `08abc3d6b` would be more useful than the standing `CHANGES_REQUESTED`, which predates every fix.

Issue #1857, opened 2026-08-22
feat(orchestrate): bounded action executor for registered local leaves


rsalus:
## Re-charter (2026-08-27 — semantic workflow plane, #1875)

The engine this issue built — intent compile, per-leaf admission, emission verification, claim-based replay — becomes the internal core of `settle`. Slice 3 (closeout segments) merged as PR #1873 (`1c0dbfe35`); slice 4 (`synthesis-closeout` incl. `create_pr`) merged as PR #1878 (`92a03fcee`, 2026-08-28), closing residue #1874; residue #1880 remains live; #1879 closed 2026-08-28 by PR #1881 (`dfb78819d`). **The suspension slice is canceled as specified**: its M3 blocker dissolves — no mid-execution suspension exists under the plane; host obligations surface at settlement. Re-scope the remaining slices against #1875 before dispatch. The `native:`/`tool:'none'` compile refusals remain the adapter seam.

---

## Summary

Runtime slice of #1856. Add a bounded executor under `exarchos_orchestrate`: one request names an intent; Exarchos compiles and executes the closed segment of registered local leaves; the operation stops only at completion, failure, or a host-owned obligation.

This is not a generic batch action. The caller never supplies an arbitrary primitive list. The executable plan comes from the registered/compiled action definition and must lower into #1258.

## Status (2026-08-28)

The first slice is delivered — PR #1872, merged `eb6abcf87`. `exarchos_orchestrate.execute_intent` ships with one compilable intent (`task-completion`), the `orchestrate.intent_executed` catalog event, claim-based commit/replay under a caller operationId (in-process same-operation serialization; cross-process racers commit-serialized with honest refusals), per-leaf derived operation identities, enforced runbook `onFail` (emission-contract violations halt even advisory leaves in block mode), a declared economy budget with a pinning summarizer, and an `operationId` filter on `event.query` for trace retrieval.

[...]

PR #1862, opened 2026-08-23
fix(gates): make the phase-gate enforcement edges actually enforce

rsalus:
## Summary

The phase-gate architecture describes four enforcement edges. Measured against the live tree at `529348175`, **two of them enforce nothing** and a third is satisfiable by a hand-written event — so the obligations the declaration surfaces record are not backed by anything at runtime. This is wave 1 of 4; it makes the edges real before later waves enrich what they carry.

Spec: [`2026-08-22-phase-gate-agnosticism.md`](https://github.com/lvlup-sw/docs/blob/main/exarchos/docs/specs/2026-08-22-phase-gate-agnosticism.md) · study: #1849

### The measurements

| Edge | Was | Now |
|---|---|---|
| runbook chain `onFail` | 71 declarations, **no executor**, none possible at that seam | advisory contract stated at the type and on the payload; `retry` (0 declarations) removed |
| `task_complete` admission | two paths to one obligation, **only one role-gated** | both hold one capability bar; runner-minted rows admit on resolved provenance |
| phase boundary | working as described | unchanged |
| config severity | `withConfigSeverity` had **zero production callers** — a disabled gate still ran and still blocked | wired, with the disabled outcome written by the producer a real run uses |

## Changes

- **`task_complete` admits on provenance, not on a name.** The caller-evidence path already required a transport-derived operator role, and says so: *"a delegated agent … cannot clear this bar no matter what it puts in `args.evidence`."* The event path required nothing, and `exarchos_event.append` is `roles: ROLE_ANY` — so that agent cleared the bar by appending instead. A row is now admitted either because it cites an `admission.evidence-recorded` id that **resolves** to a passing record bound to this task *and* this gate, or because it meets the same operator bar. Resolution, not presence: an invented id proves nothing. The gate binding is recovered by inverting the provider registry the runner already used, so a passing `check_test_adequacy` record can no longer discharge `static-analysis`. A row naming no task now discharges no task — previously an empty `details` discharged **every** task on the stream.
- **`selectSubject`'s stated invariant made true.** It documents that per-task ladder runs are always task subjects but produced one only when given a `taskId`, and no runbook step supplied one. The four per-task ladder steps now thread it, so ladder evidence is attributable per task.
- **Scope failures leave a trace.** Previously every precondition failure returned before any append, making "could not be scoped" indistinguishable from "never invoked".

The `gate.executed` event type stays **unreserved** on purpose: the shepherd skill documents an agent appending it at `layer: "CI"`, and that stays legal — it simply stops discharging a per-task obligation it was never about.

## Test Plan

- [x] `npm run typecheck` clean (root + conformance + tests)
- [x] `tests/unit/tasks` **79/79** (22 were red mid-change; the one pinning the removed behaviour, `…ProjectWideStaticAnalysis_AcceptsNoTaskId`, is **inverted**, not re-seeded)
- [x] New: `tests/integration/task-complete-admission-provenance.test.ts` (kill-probed twice — weakening to name-only reds 5 tests; weakening resolution to presence reds 2)
- [x] `tests/unit/runbooks` 133/133 · `tests/unit/verbs/gates` 646 · `tests/architecture` 230/232
- [x] `npm run render:guard` clean
- [x] Merge-suite failure **sets** diffed before/after: 26 ≡ 26, zero new, zero vanished — the documented fresh-install baseline, not this change

**Known environmental (both named in #1849):** `published-contract` (`dist/` unbuilt in a worktree) and `worktree-inventory` (accumulation).

### Behavioural change worth calling out

`deriveMcpCallerIdentity` yields `role: 'agent'`; `deriveLocalOperatorIdentity` yields `'operator'`. Manual gate satisfaction by hand-appending `gate.executed` therefore survives **over the local CLI only**, not over MCP. That is the same property the caller-evidence path always had — the consistency this PR exists to create — but it is a live contract change for anyone doing it in an MCP session.

### Added after the first CI run

CI found three things no local tier did, all of them this change's own collateral:

- The **gate-production ownership census** keys on exact `(file, kind, count)` dispositions so a new emitter fails closed. This adds two — the scope-failure trace and the not-applicable recorder — each declared with its rationale. The scope-failure entry states its compromise plainly: it records `verdict: indeterminate` in details while the wire still carries `passed: false`, because the event's verdict field has not been widened yet.
- Editing one line of the dogfood skill broke **two separately-pinned copies** of it — a vitest snapshot and a byte-identical fixture baseline. Both updated by substituting the single line, so neither loses the history it exists to hold.
- The **merge-detour saga** reached its subject by hand-appending a passing `gate.executed` row, having already worked around an earlier bypass closure the same way. It now seeds its actual subject — a worktree-bearing `task.completed` — exactly as the steps above it already seed `task.assigned`. The assertion is unchanged. Note what stays impossible: `admission.evidence-recorded` is **reserved** against untrusted append, so the proof cannot be seeded even though the signal can.


rsalus:
Closing as superseded. Premise-audited against `7f5ce233b` before deciding, and two of the four edges this PR set out to make real have been settled independently since it was opened on 2026-08-23:

| Edge | Claim here | Live tree now |
|---|---|---|
| runbook chain `onFail` | 71 declarations, **no executor**, none possible at that seam | **Stale.** The #1857 executor consumes it: `verbs/execute/executor.ts:713` reads `leaf.onFail === 'continue'` to choose `advisory-failed` over `failed`, and `verbs/execute/compile.ts:239` handles `onFail: 'retry'` — the arm this PR proposed to delete as having zero declarations. |
| `task_complete` admission | event path role-gated by nothing | **Still open**, but the fix here is the one main's own in-file guidance now rules out (see below). Refiled as #1886. |
| `withConfigSeverity` | zero production callers; a disabled gate still runs and still blocks | **Still true** — every reference is the definition, a `{@link}` beside it, or its own test. Refiled as #1886. |
| `selectSubject` per-task subjects | invariant stated but not made true; no runbook step supplied a `taskId` | **Resolved on main.** All three ladder handlers thread it — `static-analysis.ts:122`, `test-adequacy-handler.ts:223`, `mock-boundary-handler.ts:234`. |

The `task_complete` piece is the reason a rebase would not have salvaged this. `verbs/tasks/tools.ts` now carries an explicit instruction at the admission site:

> Read one event type, one shape — do NOT teach this reader to also accept the proof record; the fix belongs at the producer.

This PR's change is precisely that: admitting a row because it cites an `admission.evidence-recorded` id that resolves. The hole is real, but the patch is aimed where main has since decided the fix does not belong.

The collision is also mechanical, not just conceptual. Main rewrote the same files under the branch — `registry/actions/orchestrate/gates.ts` +303/-100 and `events/schemas.ts` +176/-29 across #1867, #1881 and #1882 — against this PR's +248 in `gate-runner.ts` and +102 in `tasks/tools.ts`.

Nothing measured here is lost: the two still-live defects are #1886 with their evidence, the analysis stays in the `2026-08-22-phase-gate-agnosticism` spec, and the study lands via #1849. The branch is preserved if any of the four integration suites are worth recovering.

PR #1863, opened 2026-08-23
feat(gates): resolve a gate's result carrier and base branch, not just its command

rsalus:
## Summary

Exarchos governs *other* repositories, but a measurable part of the gate surface was written for this one. The toolchain registry is properly polyglot — 12 toolchains, each declaring its own test/typecheck/lint/mutation commands — and the **command** axis mostly routed through it. The **result** axis had no seam at all, so gates that resolved a command correctly then read its output as if every repo were this one.

Stacked on #1862. Spec: [`2026-08-22-phase-gate-agnosticism.md`](https://github.com/lvlup-sw/docs/blob/main/exarchos/docs/specs/2026-08-22-phase-gate-agnosticism.md) · study: #1849

### Measured at `529348175`

- `check_integration_suite` — the **blocking** high-tier ladder gate — appended vitest's `--reporter=json` to whatever the registry returned, on **all 12 of 12** toolchains (`cargo test --reporter=json`, `pytest --reporter=json`), parsed vitest JSON only, and failed **closed** on any other shape. No-toolchain fallback: `npm run test:run` — this repository's own script name, not a convention. **The top ladder rung was unsatisfiable on 11 of 12 declared toolchains.**
- **10 of 40** gate modules defaulted the diff base to the literal `'main'`; **zero** detected the repository's default branch, though the provider seam exposes it and two near-verbatim private detectors already existed, both unexported.
- `check_static_analysis` — first rung of all 8 ladder cells — re-derived commands in a hard-coded 4-arm switch; `SUPPORTED_TOOLCHAINS` covered 4 of 12 and had **zero references outside its own module**.
- `verify_worktree` decided membership with a `.worktrees/` path substring — **false for the harness-native `.claude/worktrees/` layout**, so it aborted inside the very worktree this was written in.
- Verified by seeding a violation: **no guard anywhere** failed when a gate hard-coded a toolchain command.

## Changes

**The carrier seam.** `TEST_REPORT_FORMAT` mirrors the `MUTATION_DIFF_SCOPE` table beside it: keyed by toolchain id, returning a tagged descriptor so consumers branch on a kind and never on an identity, with `satisfies Record<ToolchainId, …>` making an unmapped toolchain a compile error. **Three arms only** — no `junit-xml`/`tap`, because nothing parses them. `exit-code-only` is the majority answer and a complete one.

[...]

PR #1865, opened 2026-08-23
refactor(gates): triage the gate population before instrumenting it (DR-9)

rsalus:
## Summary

Wave B of the phase-gate spec (tasks 016-022, DR-9): triage the gate population before instrumenting it.

Eight gate actions declared an obligation no chain asked them to discharge, or discharged it by reading something nothing writes. A declaration like that is worse than an absence — it inflates every census taken over "the gate population" and reads as coverage to anyone auditing the surface.

| retired | disposition |
|---|---|
| `check_convergence` + the `convergence` view | Attributed results by reading `details.dimension`, which the durable gate runner never stamps — D2 could never converge and the aggregate verdict was unsatisfiable on every automated path. **Declared INV-2 contract change.** `gate.dimension` itself is KEPT; only the consumer retires. |
| `check_coverage_thresholds` | Read Istanbul/Jest `coverage-summary.json` shapes directly — the toolchain literal this spec exists to remove — and nothing reached it. |
| `check_design_completeness` | A self-declared deprecated alias delegating to `check_plan_coverage`. Deleted outright, so the plan-structure binding has one name for the obligation instead of two. |
| `check_context_economy`, `check_operational_resilience`, `check_workflow_determinism` | Measured to be exactly the three `getDiff` callers among the gates, and structurally identical. Consolidated into **`check_diff_hygiene`**: one scanner with a rule pack, one durable row per rule under that rule's own gate name and dimension — so the D3-D5 axes they were supposed to produce are now actually produced. |
| `pre_synthesis_check` | A blocking gate nothing invoked. Folded INTO `prepare_synthesis`, carrying its resolver and provider seams across. |
| `debug_review_gate` | Held the last test-runner literal in the gate population; no chain reached it. Its coverage obligation survives as a review checklist item in the debug thorough track. |

Net action count **124 → 116**, fully closed.

## Changes

**Three new guards, each with a stated denominator**
- `gate-reachability` derives BOTH populations from live imports and compares the orphan roster for EQUALITY, so it can neither absorb a new orphan nor keep naming a fixed one.
- `retired-public-names` holds the retirement as DATA and checks five published surfaces: registry, tool-description advertisement, the generated addressing set, the projection name space, and the `--follow` set.
- `dimension-severity-path` pins the config path a dimension is nameable on.

**Recording the retirement, rather than relaxing what detects it**
- The test-inventory relocation ledger could say where a test *went* but not that one was *retired*, so nine deliberate deletions reached the reconciliation looking exactly like suites silently dropped by a stale glob. A retirement register is consulted instead — and is itself held to two conditions: a row whose file is tracked again FAILS as stale, and a row no relocation chain ends at FAILS as inert. A deletion absent from the register is still reported.
- Two files entered `@oracle-sources` scope. `diff-hygiene` names its two real authorities; `prepare-synthesis` had none (its quantified case reached the same module twice), so its assertion now collects and names the refused workflow type instead of quantifying to a boolean.

**Regenerated / re-approved**
- The compiled contract artifacts and the authority lock: the action-id registry lost 8 ids and gained 1, so the freeze is **re-approved with the retirement recorded in its note**. ⚠️ *Reviewer: this is the human approval gesture — please confirm the recorded reason.*
- `content/` skill corpus + `rendered/` + the migration baselines.

**Also carried** — `READ_ONLY_ACTIONS` and the orchestrate `slimDescription` still named the retired trio; the review runbook keeps five steps because `check_diff_hygiene` takes the slot; four playbook strings still told an agent to run a shell gate that has not existed since the TypeScript port.

## Test Plan

Every gate below was run locally against the **merged** tree that ships:

- DR-9 type-debt, DR-12 prefix-fingerprint, DR-13 prose-lint, DR-14 cast census, DR-27 measured-premise drift — all pass.
- Cycle ratchet, module-intent, outputSchema vacuity, both CLI-derivation guards, G3 report-coupled ratchet, gate-runner ownership census, `render:guard` — all pass.
- Guard-inventory reachability, G4 effect-ledger, audit-delivery closure, the full conformance suite, DR-30 suite-invariants (20/20) — all pass.
- Self-tests: check-mutation-gate, cycle-gate, module-intent, check-type-debt — all pass.
- `npm run typecheck` clean; `tests/architecture/` 264/265 (the one failure is the known environmental `worktree-inventory` accumulation, see #1849).

**Note on coverage:** the Grep Gates job aborts at its first failing step, so 23 of its 62 steps had never executed on this stack — the job died at step 39 on #1862 and step 18 on #1863. All 23 were run locally and pass.

**Kill-probed, not just observed green:** the retirement register was mutated three ways (an inert row, a stale row, and a real deletion left unregistered) and each was caught by the intended assertion; the new generated-addressing surface was mutated by re-adding a retired id and named it.


rsalus:
Closing as superseded, together with the rest of the phase-gate stack (#1862, #1863, #1865, #1866).

The stack forked at `5b3ad0dbb` on 2026-08-22. Eight commits have landed on main since, and they rewrote the surface these PRs edit rather than sitting beside it — `registry/actions/orchestrate/gates.ts` +303/-100 and `events/schemas.ts` +176/-29 across #1867 (action-contract closure), #1881 (ten gate actions paying declared evidence) and #1882. 74 of the files this stack touches were changed by that advance, including `contract-authority.lock.json` and the generated CLI, reachability and proof-fixture artifacts.

That makes rebasing the wrong instrument. #1865 in particular retires eight gate actions and re-derives a 124 → 116 count out of the exact file main has since grown, so the retirement set, the pinned cardinalities and the guard denominators all need re-deriving against the current tree, not replaying. #1866 also states in its own description that tasks 027/028/030 are blocked on #1858.

The full premise audit is on #1862. Two of the four defects the stack was built to fix are still live and have been refiled with their evidence as #1886; the other two were resolved independently on main. The `2026-08-22-phase-gate-agnosticism` spec keeps the analysis, and the measurement study lands via #1849.

Branches are preserved — reopen or cherry-pick if a specific suite here is worth recovering.

PR #1866, opened 2026-08-23
feat(gates): indeterminate reaches the wire (DR-11, tasks 024-025)

rsalus:
## Summary

Wave C of the phase-gate spec, DR-11: **indeterminate reaches the wire.** Tasks 024 and 025 of 8. Draft — 026 is next, and 027/028/030 are genuinely blocked on #1858.

`passed: boolean` had to answer four questions with two values. A gate that ran and failed, a probe that could not run, and an obligation that was withdrawn all reached the stream as `passed: false` — or, for a withdrawn one, as `passed: true`, which reads as a gate that ran and succeeded. A reader gating on the boolean cannot tell any of them apart, so a stop it should have made and a stop it should not have made look identical on the wire.

## Changes

**Task 024 — the record gains somewhere to say it did not decide.**
`GateExecutedData` gains `verdict: 'pass' | 'fail' | 'indeterminate' | 'not-applicable'`. `passed` is **retained and derived**, so every reader written against the boolean keeps working and every row already on a stream stays valid. The field is optional solely because historical rows predate it, and deliberately **not defaulted** — a default would assert a verdict the emitter never computed.

The derivation is enforced at the **schema**, not left to writers. A row where `passed` and `verdict` disagree has no defensible reading — one of them is wrong and nothing downstream can tell which — so refusing it makes the derivation a property of the record rather than a convention every emitter is trusted to have followed. It also closes the one encoding a withdrawn obligation used to have: `passed: true, verdict: 'not-applicable'` is now rejected outright.

**Task 025 — the emitter takes the verdict, and the runner stops collapsing it.**
`emitGateEvent` takes the verdict; `passed` is computed at the single place a row is minted. That is what makes the migration exhaustive — **a bare boolean stops type-checking**, so the compiler named all 20 call sites rather than anyone having to remember them.

`gate-runner.ts` was the sharper case. It is the one place that *computes* a verdict and it was the place that threw it away: the row carried the collapsed boolean at top level and the verdict only in `details`, where the readers that gate on outcomes do not look. It mints the row whole now. `details.verdict` is deliberately retained — every row written before the widening carries it only there, so a reader folding history has to keep looking there.

All 20 sites are **behaviour-preserving** (`X` → `X ? 'pass' : 'fail'`). Deciding which should instead say `indeterminate` or `not-applicable` is skip polarity — that is task 026, not something to smuggle in here.

## Test Plan

- **Additive, quantified not sampled.** `HistoricalRow_AnyShape_RemainsValid` drives fast-check over the pre-existing field space; `Passed_EqualsVerdictIsPass` runs the derivation over the *whole* vocabulary, both directions, so a fifth verdict added later inherits the case instead of slipping past a hand-listed three.
- **Replay.** Three separately maintained projections are driven over the same events twice — once in today's shape, once with the derived verdict — and required to be indistinguishable, including for a stream that **interleaves both shapes** as a real one does mid-migration. It carries its own kill probe: flipping every `passed` must move a projection, or the comparison is only proving that two identical computations agree.
- **No call site passes a bare boolean**, asserted both ways: an AST scan of every call site, *and* the declaration still typing the parameter as `GateVerdict` — because a signature that quietly went back to `boolean` would leave a call-site scan passing over sites that had all been rewritten.
- **The diff-hygiene characterization is lifted, not rewritten.** Its baseline captured call arguments while the verdict argument was still a boolean; those recorded booleans are lifted through the same derivation the emitter applies, and *only* that argument moves. The capture is evidence of what the three retired handlers did, and editing evidence to agree with new code is how a characterization quietly stops characterizing anything.
- Local: `typecheck` clean; 1332 unit tests across the gate blast radius; integration, DR-30 suite-invariants and the architecture tier green apart from the two documented environmental reds (`worktree-inventory` accumulation, `published-contract` needing a built `dist/`).

### DIFF
--- a/src/events/schemas.ts
+++ b/src/events/schemas.ts
@@ -830,13 +830,55 @@ export const GateExecutedDetailsSchema = z.object({
   promptVersion: z.string().optional(),
 }).passthrough();
 
-export const GateExecutedData = z.object({
-  gateName: z.string(),
-  layer: z.string(),
-  passed: z.boolean(),
-  duration: z.number().optional(),
-  details: z.record(z.string(), z.unknown()).optional(),
-});
+/**
+ * What a gate concluded.
+ *
+ * A boolean cannot express the two outcomes that are not a decision. A probe
+ * that could not RUN and an obligation that was WITHDRAWN both had to be
+ * written down as `passed: true` or `passed: false`, and a reader gating on
+ * the boolean could not tell either of them from a gate that actually ran and
+ * reached that answer. `indeterminate` is "no claim was decided";
+ * `not-applicable` is "the obligation was withdrawn, so there was nothing to
+ * decide". Neither is a pass, and neither is a failure.
+ */
+export const GateVerdictSchema = z.enum(['pass', 'fail', 'indeterminate', 'not-applicable']);
+export type GateVerdict = z.infer<typeof GateVerdictSchema>;
+
+export const GateExecutedData = z
+  .object({
+    gateName: z.string(),
+    layer: z.string(),
+    /**
+     * RETAINED and DERIVED — `passed === (verdict === 'pass')`. Every reader
+     * written against the boolean keeps working, and every row already on a
+     * stream stays valid: this widening is additive, not a migration.
+     */
+    passed: z.boolean(),
+    /**
+     * Optional ONLY because historical rows predate the field. A row that
+     * CARRIES it may not disagree with `passed` — see the refinement below.
+     */
+    verdict: GateVerdictSchema.optional(),
+    duration: z.number().optional(),
+    details: z.record(z.string(), z.unknown()).optional(),
+  })
+  .superRefine((row, ctx) => {
+    // Absent on a historical row, and absence is not disagreement.
+    if (row.verdict === undefined) return;
+    if (row.passed === (row.verdict === 'pass')) return;
+    // Enforced at the SCHEMA rather than left to each writer: `passed` is the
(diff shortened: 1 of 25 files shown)


rsalus:
Closing as superseded, together with the rest of the phase-gate stack (#1862, #1863, #1865, #1866).

The stack forked at `5b3ad0dbb` on 2026-08-22. Eight commits have landed on main since, and they rewrote the surface these PRs edit rather than sitting beside it — `registry/actions/orchestrate/gates.ts` +303/-100 and `events/schemas.ts` +176/-29 across #1867 (action-contract closure), #1881 (ten gate actions paying declared evidence) and #1882. 74 of the files this stack touches were changed by that advance, including `contract-authority.lock.json` and the generated CLI, reachability and proof-fixture artifacts.

That makes rebasing the wrong instrument. #1865 in particular retires eight gate actions and re-derives a 124 → 116 count out of the exact file main has since grown, so the retirement set, the pinned cardinalities and the guard denominators all need re-deriving against the current tree, not replaying. #1866 also states in its own description that tasks 027/028/030 are blocked on #1858.

The full premise audit is on #1862. Two of the four defects the stack was built to fix are still live and have been refiled with their evidence as #1886; the other two were resolved independently on main. The `2026-08-22-phase-gate-agnosticism` spec keeps the analysis, and the measurement study lands via #1849.

Branches are preserved — reopen or cherry-pick if a specific suite here is worth recovering.

Issue #1855, opened 2026-08-22
bug: PROJECTION_DEGRADED refuses workflow get/convergence while the stream is still appending


rsalus:
## Summary

`exarchos_workflow get` and several `exarchos_view` reads (`workflow_status`, `convergence`) **hard-refuse** with `PROJECTION_DEGRADED` when the fold cursor is behind the durable event tail — including lag of **1 event**. On a live `delegate` stream that is still appending gates and merges, that lag is the common case, so the canonical read path fails while the workflow is actively emitting.

The error's `suggestedFix` is often `exarchos_view` / `workflow_status`. That is the same call that just failed. The durable log (`exarchos_event query`) still answers; a later retry of `workflow_status` can succeed after a query side-effect re-folds. Callers should not have to bounce through a different tool, and the recovery tool must not refuse with the same code.

This is **not** DKG corpus-projection work (#1748). It is the live workflow-state fold refusing to serve MCP reads.

## Reproduction

Workflow: `closed-surface-first-six` (feature, phase `delegate`, hot merge/gate stream). Repo: `lvlup-sw/timedat`. Observed 2026-08-22.

### 1. `workflow get` withholds while lag is large

`exarchos_workflow get featureId=closed-surface-first-six`

Error (representative): projections stop 40 events short of the durable tail (tail 120, worst cursor 80). `PROJECTION_DEGRADED`. Suggested fix: `exarchos_view` `workflow_status`.

### 2. `workflow_status` can re-fold, then `convergence` fails on lag 1

After a successful `workflow_status`, immediately:

`exarchos_view action=convergence workflowId=closed-surface-first-six`

Error: lag **1** (tail 161, cursor 160), `staleViews: ["workflow-state"]`. Same `PROJECTION_DEGRADED` withhold.

### 3. `workflow_status` itself refuses; suggestedFix points at itself

Later in the same session (stream still appending `gate.executed`):

`exarchos_view action=workflow_status workflowId=closed-surface-first-six`

```
PROJECTION_DEGRADED
Refusing to answer from stream 'closed-surface-first-six': its projections stop 6 event(s) short of the durable tail (tail 167, worst cursor 161).
This is NOT "no data" — the answer is withheld because the fold it would be derived from provably does not cover the event tail.
Re-read the stream through exarchos_view to re-fold and clear this state.
suggestedFix: { tool: "exarchos_view", params: { action: "workflow_status", workflowId: "closed-surface-first-six" } }
[...]

Issue #1109, opened 2026-04-16
v2.8–v3.0 Cross-Cutting: Event-sourcing integrity + MCP parity + basileus-forward


rsalus:
> **v3 ADR alignment:** this issue now targets [`docs/adrs/ontological-data-fabric.md`](https://github.com/lvlup-sw/basileus/blob/main/docs/adrs/ontological-data-fabric.md) §3.6 capability resolution (handshake-authoritative) + §4.6 ontology-version cache invalidation. Prior references to the superseded v1/v2 coordination ADR and the 2026-04-19 ingestion/retrieval design docs were retargeted in-place on 2026-04-20; the v3 ADR inlines and supersedes all three (§19).

## Overview

Cross-cutting constraints that apply to **every v3.0 pillar**. These are not optional — they are design invariants.

## Constraint 1: Event-Sourcing Integrity

Every new command surface must answer:
1. **Does it read from the event store?** (projections)
2. **Does it write to the event store?** (emit events)
3. **Does it stream from the event store?** (subscriptions)
4. **Can its output be reconstructed from events alone?** (event-sourcing guarantee)

No new state management paths. Lifecycle verbs are windows into the event store.

**Events emitted by new commands:**
- `diagnostic.executed` (doctor)
- `wait.started`, `wait.completed`, `wait.timeout` (wait)
- `export.created` (export)
- `init.executed` (init)

## Constraint 2: MCP Parity

Every output shape must work identically from CLI and MCP facades. The HATEOAS envelope is the same whether returned from `exarchos describe --json` or `exarchos_orchestrate({ action: "describe" })`.

**Implementation:** All commands route through the shared dispatch core. CLI and MCP adapters are thin presentation layers.

## Constraint 3: Basileus-Forward

No design decisions that make MCP second-class. The planned Basileus integration uses MCP as the transport protocol for distributed agent coordination:
- CLI = local human/agent interface (single machine, single session)
- MCP = inter-agent coordination protocol (local ↔ remote)
- Both facades remain first-class

**Authoritative design (2026-04-18):** [basileus: ADR Exarchos ↔ Basileus Coordination Architecture](https://github.com/lvlup-sw/basileus/blob/main/docs/adrs/ontological-data-fabric.md)

The ADR cements Basileus-forward into concrete invariants:

[...]

PR #1867, opened 2026-08-23
feat: Fail-close ActionId admission on trusted caller facts (rebased + #1855-aware)

rsalus:
## Summary

This PR is the rebase-and-synthesize of #1858 onto current `main` (which now includes #1855/#1859's fold-at-tail fix). It preserves every behavioral change from #1858 while keeping the projection fold-at-tail seam that #1855 added.

## Why a new PR

PR #1858 was developed against pre-#1855 code. Its tree reverses the projection changes that #1855 landed — it removes `materializeAt`/`discardFold`, inlines `isInternalSentinelStream`, drops `fold-at-tail.ts`, and reverts 19 view handlers back to a `queryDeltaEvents` + `materialize` direct pattern. The two PRs have *opposite* semantics for `PROJECTION_DEGRADED`:

- #1855: fold forward, never expose stale; refuse only when coverage is genuinely undecidable
- #1858: drop stale payload, refuse, require callers to read `_meta`

A mechanical merge produces 162 conflicting files, and a naive "apply the PR's tree on top of `origin/main`" produces broken TypeScript (4 errors in `src/projections/fold-at-tail.ts`) because the PR depends on symbols that #1855 removed.

The action-admission logic in #1858 — the actual subject of the PR — does not depend on either side's projection implementation. It reads HSM facts via `eventStore.query()` + the in-memory `workflowStateProjection.apply()`, not through the materializer or the fold-at-tail seam. So the synthesis keeps #1855's projection system intact and adds only the small PR-specific surface that needs to coexist with it.

## Changes

- **Squashed PR tree onto `origin/main` (`5b3ad0dbb`)** — the full action-contract admission machinery lands as a single commit (`7f63e6473`).
- **Revert all of the PR's `src/projections/` reverts** — `materializeAt`/`discardFold`/exported `isInternalSentinelStream` come back; `fold-at-tail.ts` and the 19 view handlers are origin/main's versions.
- **Add the small PR-specific surface that has to coexist with #1855**:
  - `guardProjectionDegraded` and `resolveProjectionStreamId` on `degraded-result.ts` (the consumer-side guard, separate from the fold seam)
  - `deriveAdvertised` on the next-action reducer (the allow-only ActionIds envelope, distinct from `derive`'s HSM/control envelope)
- **Route 7 PR-era call sites back through `foldToTail`** in `src/verbs/gates/` and `src/workflow/handlers/` — the fold-seam architecture guard was red on these.
- **Regenerate audit baselines** for the new test population: `test-inventory-baseline.json` (+157 lines), `reference-census.json`, `guard-liveness-baseline.json`. Bump `tools/audit` locality exemption 52 → 56.
- **Revert four `src/workflow/` files** (`composite.ts`, `handlers/shared.ts`, `state-store.ts`, `tools.ts`) to `origin/main` — they were collateral PR changes that depended on the broken seam.

## Cross-cutting (#1109) verification

- [x] **Event-sourcing integrity:** admission HSM facts fold `workflowStateProjection` over `eventStore.query(featureId)`. No new existence signal from `.state.json`. Events: existing `diagnostic.executed` / contract emissions; no new event types.
- [x] **MCP parity:** CLI and MCP share `dispatch`. Default process resolver is wired from both `createCliDispatchContext` and `createServer`.
- [x] **Basileus-forward:** no new config file; process grant is posture + optional env kill switch for cache hints.
- [x] **Capability resolution:** ActionId needs use the trusted authorization snapshot. Handshake `list()` is not the need set.

## Backend-quality dimensions

DIM-3 Contracts, DIM-2 Observability, DIM-6 Architecture.

## Test Plan

- [x] `npm run typecheck` clean
- [x] `npm run test:run` — 1123 passed / 1 failed / 39 skipped (the one failure is `WorktreeInventory_EveryRegisteredWorktree_IsRecorded`, which is `skipIf(liveWorktreeCount() <= 1)` — runs only on a developer host with multiple worktrees; my host has 67, the committed inventory says 95, so this is environmental)
- [ ] `cd servers/exarchos-mcp && npm run test:run` — N/A
- [ ] `npm run skills:guard` — N/A (no `skills/` touched in this PR)

## Commits

```
f4d0bb45d fix(synthesis): route fold-using call sites through foldToTail; regenerate baselines
9a744d0ad feat(projections): port action-contract closure's projection guard without regressing #1855
7f63e6473 feat: Fail-close ActionId admission on trusted caller facts
5b3ad0dbb fix(projections): fold to the tail instead of refusing to answer (#1855) (#1859)  ← origin/main
```

## Related

- Supersedes #1858. All behavioral changes from #1858 are preserved; the projection conflict has been resolved by reverting the projection reverts and adding only the necessary coexistence surface.
- #1855, #1859 — the fold-at-tail fix that #1858 was originally developed against and which the synthesis preserves.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


rsalus:
## CI repair — the projection conflict resolved the way this PR declared

Pushed `416158a` + `e05eeb5`. All eleven test files the red lanes reported are green locally.

### What was actually wrong

The PR body says the synthesis "keeps #1855's projection system intact". The **source tree** does. A set of **tests** did not — they came across as #1858's pre-#1855 versions, asserting the contract #1855 replaced. The follow-up commit `98089fdb5` then made those tests pass by wiring #1858's refuse-on-marker semantics *back into two handlers*, which regressed #1855 in the shipped code and broke a different set of guards.

The two contracts are exact inverses on identical injected data (`tail 42, cursor 13, lag 29`):

| | `Governance_Dr4_DegradedProjection_IsNeverServedAsSuccess` (added by this PR) | `FoldAtTail_FabricatedDegradedMarker_DoesNotWedgeAHealthyStream` (on `main`) |
|---|---|---|
| verdict | `get` must **refuse** | `get` must **succeed** |

They cannot both hold. Per the PR's own stated resolution, **#1855 wins**.

### Source fixes

- **`src/workflow/handlers/get.ts`** — dropped the `guardProjectionDegraded` call. It refused any read of a stream carrying a durable `projection.degraded` row, which is precisely the wedge #1855 removed. Its `console.warn` was also the sole `NoConsoleInProduction` violation.
- **`src/verbs/composite.ts`** — dropped `PROJECTION_DERIVED_ORCHESTRATE_ACTIONS` and the same guard over it; the four readiness/reliability verbs were being wedged by a spent observation, against `Consumer_StaleFoldAndDurableMarker_IsNotWedged`.
  Separately: **restored `foldToTail` in `adaptSetupWorktree`** — the missed *eighth* call site of the seven this PR routed back, and with it the coverage-failure arm that keeps a managed worktree off a stale `main`.
- **`src/workflow/tools.ts`** — removed the `configureWorkflowMaterializer` no-op shim. It broke two architecture guards at once: the file is a published re-export barrel, and that symbol is the reachable-controls **kill fixture** — the shipped tree using it *is* the defect that guard exists to demonstrate.
- **`src/architecture/layer-boundaries-seam.ts`** — admitted the new declaration edges: `config`, `contract`, `dispatch`, `adapters/mcp` → `registry`, and `contract` → `workflow`. An action contract is DECLARED once under `registry/` and the MCP registration / CLI surface / describe projection / compiled contract are projections of it; `contract` reaches `workflow` for the admission verdict, which is the HSM's to give. Each row states its reason; the dispatch row's stated width moves 17 → 18.

### Test fixes

Fifteen files reverted to `origin/main` — every one asserting a contract that no longer ships:

- `freshness` / `degraded-consumers` imported the **removed** `assessStreamFreshness`
- `telemetry-queries`, `check-convergence`, `check-event-emissions`, `context-economy`, `prepare-delegation` stubbed `materialize` under sources that fold through `foldToTail` — so the stub covered a seam nothing calls and every verdict came from an **empty fold** (`context-economy` read 0 session tokens instead of 5000; `prepare-delegation` mistranslated a fail-closed `PHASE_BLOCKED` into `PREPARE_DELEGATION_FAILED`)
- seven `workflow/` files called the retired materializer setter

The pre-#1855 `Governance_Dr4_...` case went with them; the claim that replaced it is asserted at the seam in `fold-at-tail.test.ts`.

Ten new test files now declare `@oracle-sources`. Scope there is decided by assertion **shape**, so a file comparing two computed values owes two independent authorities — each declaration names them and says why they are not one authority wearing two names.

### On the `verbs/merge/merge-orchestrate.*` failures

If you run the `core` project locally you will see 22 of them. **They are environmental, not a regression.** They fail when `main` is live in a *sibling* worktree, so preflight refuses with `Target branch 'main' is checked out in sibling worktree`. I baselined `origin/main` from an equivalent sibling worktree: it fails them identically (25 there vs 22 here). CI uses a single checkout and reports them green on this PR.

### Verification

- `npx tsc --noEmit` — clean
- `npx vitest run --project core` — 12,693 passed; only the six environmental `merge-orchestrate` files above
- `npx vitest run --project unit` — 1,123 passed, 1 host-environmental failure (`WorktreeInventory`, `skipIf(liveWorktreeCount() <= 1)`; this host has 67)

One thing worth flagging for review rather than fixing here: `npm run test:run` runs the `unit` project **only**, which is why the `core` lane's ~28 red tests were invisible to this PR's original test plan. `npx vitest run` is the command that covers both.

rsalus:
## Round two — `7d27db4`

Both failures fixed. One of them was **not new**: it reproduces unchanged at `98089fdb5`. The Root Package job exits on its first red step, so while the `unit` project was failing, the `acceptance` project **never ran at all**. Fixing `unit` is what made it visible.

### `Validate No Legacy` — orphaned exports

Removing both call sites of `guardProjectionDegraded` left it and `resolveProjectionStreamId` exported and unreferenced, which is exactly what the knip sweep exists to catch. Its own first-listed remedy is *"DELETE it"*, so they are deleted.

The question they answered is answered earlier now: a lagging fold is folded forward before any read consults the durable marker, so there is nothing left for a consumer-side refusal to protect. `toProjectionDegradedResult` / `readProjectionDegradedState` stay — the row is still written and still readable. Nothing reads it *to refuse*.

Knock-on: the deleted resolver held one `as const`, so the DR-14 cast census dropped 1719 → 1718. That floor is **symmetric** — a paydown that isn't re-pinned leaves a stale baseline masking the next regression — so the baseline slides down by one with provenance recorded in place.

### `Root Package` — the host-owned case named a local action

`Dispatch_HostOwned_DoesNotExecuteObligation` asserts that a blocking host obligation short-circuits before the handler runs, but named **`cutover_decide`** as its subject:

```ts
executionAuthority: { kind: 'local' },   // src/registry/actions/orchestrate/cutover.ts
```

`cutover_decide` gates on operator posture *inside* its handler and **must keep running** — its own `ensures` obliges it to append the `admission.rollout-decision` fact. Short-circuiting it would hand the caller an obligation where the recorded decision belongs. So the handler ran and returned `{ ran: true }` where the test wanted `{ obligation: 'human-approval' }`.

Retargeted to **`discover_bridge`**, which genuinely declares `{ kind: 'host', obligation: 'human-approval' }`, plus the `artifact` its schema requires — schema validation runs *before* admission, so without it the dispatch fails on `INVALID_INPUT` and never reaches the check under test.

**The dispatch short-circuit itself was correct all along.** The defect was the test's choice of subject, and the mechanism is now actually exercised rather than passing over an action it never applied to.

### Coverage note

The earlier round ran two vitest projects. There are **six**, and `acceptance` is where this hid. All six now run clean:

| project | result |
|---|---|
| `unit` | 1123 passed — 1 host-environmental (`WorktreeInventory`, `skipIf(liveWorktreeCount() <= 1)`; this host has 67) |
| `core` | 12,693 passed — only the six environmental `merge-orchestrate` files |
| `acceptance` | 81 passed |
| `conformance` | 323 passed |
| `outcome` | 32 passed |
| `process` | 23 passed |

Plus `npx tsc --noEmit` clean and `tools/audit/gates/validate-no-legacy.sh` OK (32 passed, knip clean).

rsalus:
@coderabbitai approve

Issue #1875, opened 2026-08-28
Decision: adopt the semantic workflow plane (capsule + settlement) as the Interaction Economy endpoint


rsalus:
## Summary

**Decision: adopt the semantic workflow plane — the capsule-and-settlement model — as the endpoint of the Interaction Economy program (#1856).** The unit of agent-visible MCP interaction becomes the whole workflow batch: `prepare` compiles intent, bound knowledge, and authority into an immutable, content-addressed capsule; the harness executes it with zero governance calls; `settle` adjudicates one batch of claims and evidence. Two calls on the normal path, three on the exception path.

This issue is the recorded charter decision that the #1856 / #1599 amendments reference. Building against the prior M3/M4/M5 shapes after this decision is premise drift.

Design source: [the plane document](https://github.com/lvlup-sw/docs/blob/main/exarchos/docs/references/exarchos-semantic-workflow-plane.html). Reviewed impact analysis: [semantic plane transposition](https://github.com/lvlup-sw/docs/blob/main/exarchos/docs/references/exarchos-semantic-plane-transposition.html) (docs commit `a11262f`, finalized 2026-08-27).

## Rule amendments (recorded here; applied in #1856 and #1599)

1. **Fine-grained events.** The plane moves execution-phase granularity to runtime journals and keeps six semantic event kinds in the ledger. This conflicts with the pinned constraint "keep fine-grained events internal, append-only, replayable." The split is **not decided by this issue** — it has its own decision issue (see Linked issues). Until that decision, the pinned constraint stands.
2. **In-process local leaves.** The constraint "registered local leaves execute in-process" re-scopes to the leaves internal to `prepare` and `settle`. Execution-phase leaves do not run inside Exarchos.
3. **Intent-as-unit (coordination rule 7).** The capsule batch becomes the unit of agent-visible interaction. `settle(batch)` is result *submission* against a typed settlement contract — composition stays Exarchos-compiled, never caller-authored. The no-generic-batch rule is preserved in that form.
4. **IR-first sequencing (coordination rule 3).** The lowering discipline re-binds to the **kernel** (below). The plane's "extract the definition IR from proven capsules" sequencing is rejected: the IR is a cross-service surface and is designed up-front.

## #1258 re-charter resolution: kernel-first (B), with A's generation rule and C's vocabulary

[...]

Issue #1856, opened 2026-08-22
Epic: Interaction Economy — intent-level MCP composition


rsalus:
## Summary

Make **user intent**, not internal gate/event sequencing, the unit of agent-visible MCP interaction.

Exarchos already owns the durable facts, HSM guards, action registry, effect plans, and HATEOAS affordances. The missing property is closure: the runtime returns runbook steps for the agent to interpret instead of compiling and executing the closed local segment itself. This epic coordinates the existing work that closes that gap without weakening event sourcing or crossing the harness boundary.

Research and call census: [lvlup-sw/docs#7](https://github.com/lvlup-sw/docs/pull/7).

## Status (2026-08-27)

M1 is delivered. Three residues remain.

**2026-08-27 (revision): the semantic workflow plane is adopted as this epic’s endpoint — decision record #1875.** The unit of agent-visible interaction becomes the workflow batch: `prepare` compiles an immutable capsule, the harness executes it with zero governance calls, and `settle` adjudicates one batch. M3–M5 are re-chartered in place below; the load-bearing constraints carry inline amendments. Event authority is DECIDED — #1876 recorded 2026-08-29 (composed Options 1+2, staged additively; both vocabularies stay in the ledger until the preconditions land — closure guard PR #1883, bundle oracle PR #1885 and partition map PR #1887 are on `main`; the evidence-reader widening is in flight). #1876 closed 2026-09-03 as decided; the migration ordering (first bundle producer, six semantic kinds with emitters, per-event flips, deletions) is tracked in #1888. strategos#193 measured not started 2026-09-03, so the semantic kinds and settlement composition wait on it. M2 slice 3 (closeout segments) merged as PR #1873 (`1c0dbfe35`, 2026-08-27) and slice 4 (`synthesis-closeout` incl. `create_pr`) as PR #1878 (`92a03fcee`, 2026-08-28), closing #1874; residue #1879 closed 2026-08-28 by PR #1881 (`dfb78819d` — the unpaid durable-evidence class re-measured at ten actions, all repaired; `validate_pr_stack` now segment-eligible); #1880 remains live; the suspension slice is canceled as specified.

- PR #1867 landed the action contract: one canonical contract block for each of the 124 actions, one admission evaluator, and advertise/dispatch parity.
- PR #1870 (merged `3d14e0428`) landed the emission evidence for anchors 036 and 039:
  - All 124 actions declare emissions or a reasoned abstention (59 declared, 65 abstain).
[...]

PR #1878, opened 2026-08-28
feat(orchestrate): compile the synthesis-closeout segment (#1857 slice 4) + two plan gates pay their declared evidence

### LINKED ISSUE #1874: Two plan gates declare durable-evidence ensures their handlers never pay (check_task_decomposition, spec_coverage_check)
## Summary

Found while implementing #1857 slice 3 (PR #1873). Two gate actions declare `ensures: durable-evidence` and never write it:

- `check_task_decomposition` (`src/registry/actions/orchestrate/verification.ts`, registration near the top of the file)
- `spec_coverage_check` (`src/registry/actions/orchestrate/verification.ts`, further down)

Both handlers emit only a bare `gate.executed` and return success. Every postcondition-observing caller — the ordinary dispatch path's ensure verification, and the bounded intent executor's per-leaf emission check — reads a success carrier that has broken its own contract: dispatch returns `ENSURE_CONTRACT_VIOLATED`, and inside an `execute_intent` segment the leaf halts the segment regardless of its `onFail` policy.

Retracting the declarations is not available: `tests/unit/registry/action-contract-coordination.test.ts` (`Registry_GateEnsures_NameDurableEvidence`) requires every action carrying gate metadata to name a durable-evidence ensure.

PR #1873 repaired the same defect for `check_security_scan`, `check_convergence`, and `check_invariant_conformance` by routing their verdict bodies through `runPhaseGateWithEvidence` (the shared phase-gate runner) with a registered gate class each, plus `sameOperationGateKey` on the legacy self-emitted `gate.executed` so a same-operation retry leaves one row. These two gates were deliberately left out of that PR — and consequently out of the `plan-closeout` runbook, which explains in a comment why the advisory decomposition gate and the spec-coverage check are absent.

## Repair route

Mirror PR #1873's pattern for both handlers:

1. Register `task-decomposition` and `spec-coverage` gate classes in `src/verbs/gates/gate-provider-registry.ts` and the compile-time tuple in `src/events/event-registration.ts`.
2. Wrap each verdict body in `runPhaseGateWithEvidence` with an evidence subject derived from the gate's inputs.
3. Key any provider-side `gate.executed` append with `sameOperationGateKey`.
4. Regression-test on the dispatch path (see `tests/unit/verbs/gates/review-gate-evidence-dispatch.test.ts` for the shape).

Once repaired, both gates become eligible for the `plan-closeout` intent segment; extending that runbook is a separate decision (its exclusion comment should be updated either way).

Part of the #1856 interaction-economy epic's contract-integrity residue.

rsalus:
## Summary

#1857 slice 4, re-scoped against the semantic-plane re-charter (#1875): a compilable `synthesis-closeout` intent whose segment includes the registered `create_pr` leaf — proving that external credentials (the local gh/VCS provider) do not imply a second agent round-trip. Rides on the #1874 repair: both plan gates now pay the durable evidence they declare. Part of #1856 / #1857 (both stay open).

Fixes #1874

**Scoping under #1875:** the executor engine is explicitly Carried ("executor engine #1872/#1873 → the `settle` core"), so this slice extends it in place. The canceled suspension scope is not re-introduced. `CompiledSegment` remains the documented interim form; its lowering target is re-bound from the Workflow Builder IR to the `Strategos.Contracts` kernel (lvlup-sw/strategos#193). One spec correction: the issue's `needs:vcs` phrasing names a capability that does not exist — `create_pr` declares `needs: none` because VCS provider calls are outside the closed capability vocabulary; the round-trip claim is proven structurally (the leaf compiles and executes in-process), not by a capability token.

## Changes

**#1874 — two plan gates pay their durable-evidence ensures** (`0686f4242`, `331c546c1`, `7e0db9b1b`)
- `check_task_decomposition` and `spec_coverage_check` route their verdict bodies through `runPhaseGateWithEvidence` with registered gate classes (`task-decomposition`, `spec-coverage`) in the provider registry and the compile-time event-registration tuple; the legacy self-emitted `gate.executed` is keyed with `sameOperationGateKey`.
- `spec_coverage_check` gains a required `featureId` (it previously had no stream to record evidence against) — wire-visible, so the plan skill call sites and rendered variants are updated and the generated contract artifacts re-recorded.
- Real-dispatch regression tests mirror `review-gate-evidence-dispatch` (evidence row appended, no `ENSURE_CONTRACT_VIOLATED`), plus a kill probe.
- `PLAN_CLOSEOUT`'s exclusion comment now states the true remaining obstacle (the missing `repoRoot` argument / a scope choice), not the repaired defect. Extending that segment stays a separate decision.

**Slice 4 mechanism — observation stream from the declaration** (`da7697c11`)
- `create_pr`'s durable records land on the shared `vcs` stream, not the workflow stream. `vcs` joins `INFRA_STREAM_IDS`; `create_pr` declares `{ kind: 'stream', selector: 'vcs' }`; `observationStreamId` precedence is reversed so a declared infrastructure stream beats the arg-derived one (behaviorally inert for every prior declarer — none can supply a feature arg). One authority now serves both the dispatch ensure-check and the executor's per-leaf emission verification.
- `CompiledLeaf` carries `observationStreamId`; `runLeaf` uses it at all four observation sites while `tailSequence` still filters to the segment's own stream. Per-leaf receipt events now name the stream their sequences number.
- `create_pr` keys its two phase appends off the ambient operation id (derived per-leaf id inside a segment), closing a crash-retry duplicate-PR hole; the remote precheck covers fresh dispatches.

**Slice 4 intent — `synthesis-closeout`** (`209a1ad5f`, `df4e5eb86`)
- New segment runbook (mirroring the `PLAN_CLOSEOUT` precedent; `SYNTHESIS_FLOW` untouched): `validate_pr_body` with `enforce: true` (new opt-in flag — a body missing required sections now fails the leaf instead of committing as passed), then `create_pr`.
- Excluded with stated reasons: `prepare_synthesis` (shells out; it is the readiness verdict owed before the closeout), `validate_pr_stack` (declares durable evidence it never pays — the third member of the #1874 class, filed as residue), and the `state.patched` PR-URL patch (an obligation left with the caller; the runbook, arg-schema, and tool description all disclose it).
- Typed `SynthesisCloseoutArgs` in `INTENT_ARG_SCHEMAS`; `execute_intent` description names the fourth intent inside its 280-token budget; economy budget entry added; compile now refuses a step absent from the executor's handler table (`INTENT_NOT_CLOSED`) before any effect.
- Test triple: compile, executor, emission-scope, plus a composition-parity suite over `create_pr`'s journal rows (provider stubbed statefully so the crash-retry precheck is genuinely reached — asserts `createPr` called exactly once) and a kill probe forcing the observation stream back to the segment stream, proving the mechanism is load-bearing.
- Pinned surfaces re-recorded: tools-list golden, cli-surface, proof-fixtures, dev-catalog, runbook count/shape pins, test-inventory baseline (surgical update, relocation ledger preserved).

## Test Plan

- `npm run typecheck` (root + conformance + tests) — clean
- `npm run test:run` — 1125 passed; the single failure (`worktree-inventory`) reproduces identically on `main` in the primary worktree (live-machine census drift; `it.skipIf` skips it on CI)
- `npm run test:conformance` — 323 passed
- Scoped: `tests/unit/verbs/execute`, `tests/unit/verbs/gates`, `tests/unit/runbooks`, `tests/unit/registry`, `tests/unit/dispatch`, `tests/unit/verbs/vcs`, `tests/unit/projections`, `tests/acceptance`, `tests/integration`, `tests/core` — all green
- Regeneration drift check: `cli-contract-seam.ts` + `compiler/generate.ts` re-run clean against the committed tree; `render:guard`, gate-runner ownership census, prose lint, description budget guard — all clean
- Adversarially reviewed by three independent lenses (guard vacuity with seeded violations, spec compliance vs #1857/#1856/#1875, emission/replay correctness); all six must-fix findings repaired in `f1fd9f271`, residue filed separately


rsalus:
Addressed in 9be074bc3: `handleExecuteIntent` now refuses any reserved infrastructure stream id (`INFRA_STREAM_IDS` via `isFeatureStream`) as the intent subject, after alias reconciliation and before `compileIntent`, so the guard covers both the `streamId` and `featureId` spellings with one check. Two regression tests assert the refusal for each alias and that nothing lands on the reserved stream.

rsalus:
@coderabbitai approve

Issue #1898, opened 2026-09-08
Tracking: event-authority migration — the remainder after #1888


rsalus:
## Summary

Successor to **#1888**, which closed 2026-09-08 on its stated condition — the first real custody move (item 2, PR #1890) and the first charter-backed flip (item 6, PR #1891) both on `main` with green oracles. That condition marked the **exit from the additive steady state**; it was never "the ordering is finished." Seven items of the decided migration are undone and this issue is their home.

Nothing here is a new proposal. Every item below is carried verbatim from #1888's ordering, which carried it from the event-authority decision (#1876, ratified 2026-08-29). Decision record: https://github.com/lvlup-sw/exarchos/issues/1876#issuecomment-5465417502 · plane decision #1875 · coordination epic #1856 · roadmap #1599.

## Landed under #1888 — not carried here

| Item | Shipped |
|---|---|
| Three-arm closure guard (precondition) | PR #1883 (`7f5ce233b`) |
| Bundle store + `runBundleIntegrityCheck` C2 oracle (precondition) | PR #1885 (`5e46f644d`) |
| Partition map + differential fold + raw-reader census (precondition) | PR #1887 (`ec5ff0285`) |
| 1. Evidence-reader widening (precondition) | PR #1889 (`e4c6b579a`) |
| 2. First bundle producer — `bundleRefs` on `intent_executed` | PR #1890 (`c29c930a5`) |
| 6. First charter-backed flips — demotion arm; backlog 16 → 9 | PR #1891 (`f5338b5e5`) |
| 7. Playbook `gate.executed` emit-instructions — by derivation, not excision | PR #1894 (`1fca81763`) |

## Remaining ordering

Independently landable, independently revertible. Each EventType addition or removal is a **six-surface** edit (type + maps, annotations — a load-time throw, not a tsc error — projection case, three `toHaveLength` pins, identifier-stability snapshot) plus emission declarations.

[...]

Issue #1599, opened 2026-06-22
Exarchos Roadmap — Master Tracker (adaptive-depth substrate → workflow IR)


rsalus:
> **`tracking` · never-closing · the single index for the Exarchos roadmap.** Walk the links here to reconstruct the program. Built 2026-06-21 from a 13-cluster comprehensive audit of every open issue (workflow run `wf_346099d2-999`, 122 findings) + an adversarial critique pass. **Closure-status synced 2026-06-23** (Z1/v2.11 fully shipped — see Status update below). Companion to the Basileus Master Tracker (`lvlup-sw/basileus#182`). The narrative companion is the **north-star system-design doc** (`docs/system-design.html`, in progress).

## North Star

**The entire backlog is one adaptive-depth governance substrate converging on a shared workflow IR.** Three zones layer onto the same primitives:

- **Z1 — Adaptive depth (v2.11): ✅ SHIPPED 2026-06-23.** verification *and* planning depth became risk-proportional on one `ResolveGateSetCtx` / `resolveGateSet` / phase-kind binding — ~~`#1515`~~ `riskTier` (verification — PR #1600 + SIV closeout #1615) + ~~`#1581`~~ `designDepth` (planning — PR #1611) + ~~`#1592`~~ additive REVIEW/SYNTHESIZE obligations (PR #1614). **Live residue:** `#1510` (retire init/install aliases) + `#1612`/`#1613` (DR-7 GitLab/ADO PR-comment harvesting).
- **Z2 — Runtime supervision (v2.12):** runtime state becomes observable through one event-sourced liveness + subscription convention — `#1090` lifecycle verbs **own** it; `#1574` worktree manager + `#1342` Marten consumers **consume** it.
- **Z3 — SDK consolidation (v3.0):** `#1258` Workflow Builder SDK collapses both into the IR, **deleting** `hsm-definitions.ts` + `playbooks.ts` and migrating built-ins (`#1253`).

**Unifying discipline:** every interim delta must be a *behavior-preserving lowering with a named IR/combinator mapping* — or it is churn the SDK will undo.

## Integration map (acyclic dependency ordering)

```
Z1  resolver:  one ResolveGateSetCtx  ◄── #1515 riskTier ─ #1581 designDepth ─ #1592 obligations   ✅ ALL SHIPPED
                                            (#1583 designDepth → #1582 collapse;  #1586 dispatch-wire → #1587/#1589   — all closed)

[...]

PR #1905, opened 2026-09-12
feat(contract): the Exarchos capsule, derived from the published kernel rather than mirrored

rsalus:
## Summary

The first slice of the semantic workflow plane: `ExarchosCapsuleV1`, the immutable
artifact `prepare` will compile and `settle` will adjudicate against. Contract only —
no action, no event type, no settlement-endpoint edit, no authority-lock re-approval.

**The gate this work waited on had already lifted.** The charter required the
workflow-definition kernel frozen in `Strategos.Contracts` before any capsule slice.
strategos#193 closed and the kernel shipped in 0.13.0/0.14.0 carrying both `authority`
and `contentHash`; this repository already pins 0.14.0. Three trackers still assert
"not started" — #1599 in three places, #1856, and #1898's gate table. Those corrections
go on the issues, not in code, and #1898 items 3 and 4 are not adjacent to this work,
they **are** it.

### The problem this contract had to solve

The kernel declares its objects open and every field optional, so `{}` satisfies
`WorkflowAuthorityV1`. This repository closes its wire contracts, and the charter calls
a hand-written mirror a defect. Three approaches were refuted **by measurement** before
anything was written:

| Approach | Why it fails |
|---|---|
| `.strict()` on the import | Closes ONE object. Measured: still accepts `{invariants:[{statement:'x', EVIL:'yes'}]}` |
| A refinement | Invisible to `z.toJSONSchema` — Ajv and Zod would disagree by construction, and the round-trip guard would read that as a defect here |
| Import the type, hand-author the schema | Every kernel field is optional, so a deliberately stale copy compiles clean |

What ships is a derivation with a proof that cannot pass vacuously: stripping **only**
`additionalProperties`, the JSON Schema emitted from `node_modules` deep-equals the one
emitted from our transform. One side is the package and the other is ours, so it is not
a copy compared against itself.

### `graph` was decided by one measurement, not by preference

The kernel does not merely describe a graph — it enforces one through a refinement, and
a refinement does not survive into JSON Schema. On a definition whose transition names a
step that does not exist:

| | |
|---|---|
| Zod rejects it | **true** — the kernel's rule works |
| `z.toJSONSchema(WorkflowDefinitionV1Schema)` | **succeeds** — the rule is silently dropped |
| Ajv accepts the same document | **true** |
| Ajv and Zod agree | **false** |

Embedding the kernel definition would import a schema that cannot satisfy this
contract's own round-trip guard. So structure stays referential and integrity is a
separate pass.

### The reference pass is necessary — verified, not assumed

The measurement above is the proof that reference rules cannot live inside a
round-tripped schema. Beyond that, the kernel's own enforcer covers only "this property
names an entry of that collection" for its own document: it sees none of the capsule's
cross-references and cannot express acyclicity at all. Where a capsule names a kernel
definition, the pass **delegates** to `WorkflowDefinitionV1Schema.safeParse` rather than
reimplementing rules the package owns.

Every rejecting fixture is a document the schema **accepts**, asserted in the same test.
A case the schema also refused would prove the pass redundant.

## Changes

- **`src/contract/capsule/exarchos-capsule.ts`** — nine top-level sections;
  `executionProfile` optional, the rest required. Every constraint is structural (a
  required key, a `.min(1)`, a discriminated union) because only structure survives the
  Ajv boundary. `knowledge` is a discriminated union rather than an enum plus a rule for
  exactly that reason.
- **`src/contract/capsule/kernel-derivation.ts`** — `deepStrictify`,
  `requireNonEmptyArrayFields`, `unwrapOptional`, `reachableZodNodeTypes`. Only the
  transform is authored; every property name, pattern and bound still flows from
  `node_modules`.
- **`src/contract/capsule/capsule-references.ts`** — nine violation kinds over task
  references, dependency acyclicity, the completion predicate's own fact and event
  declarations, settlement's required results, and step references into a supplied
  kernel definition.
- **`src/contract/capsule/exarchos-capsule-fixtures.ts`** — one corpus, in `src/`
  because BOTH validators import it. Two corpora would drift, and the drift would look
  like agreement.
- **Generated artifact + regeneration CLI + drift guard.**
- **`src/architecture/effect-ledger.ts`** — vets `@lvlup-sw/strategos-contracts` inert.
  This is its first shipped import and the census was right to fail closed until a human
  judged it. The judgement is re-measurable rather than asserted: `sideEffects: false`,
  `zod` as the only non-relative import anywhere in its emitted output, and no `node:`
  builtin, `fetch` or socket primitive in it.
- **`tools/audit/knip-allowlist.json`** — deletes the row that recorded the package as
  "pinned, deliberately not imported yet", exactly as importing it predicted. Count holds
  at 115.
- **`tests/unit/tsconfig-strictness.test.ts`** — cast baseline 1715 → 1722 with recorded
  provenance. Four of the seven predate this change; the three that are new arrived after
  a paydown inside it, twelve down to three.
- **`tools/audit/test-inventory-baseline.json`** — four rows spliced, four totals bumped.
  Not regenerated.

Both new modules carry `RESERVED` headers: their only production consumer is `settle`,
which is the next PR. An expired `RESERVED` stub fails the gate by design.

## Test Plan

- `npm run typecheck` — clean across all three projects
- `npm run lint` — exit 0
- `npx vitest run --project core` — **13,254 passed**, 0 failed
- `npx vitest run --project unit` — **1,165 passed**, 0 failed
- `npm run test:conformance` — **328 passed**
- `node tools/audit/gates/check-module-intent.mjs` — exit 0
- `bash tools/audit/gates/validate-no-legacy.sh` — OK, 115 findings / 115 entries
- Artifact regenerated twice: byte-identical, and unchanged by the schema refactor
- Ajv compiles the checked-in bytes

Kill probes, each biting precisely:

| Probe | Result |
|---|---|
| The resolver always reports sound | 19 of 36 red — the 17 survivors are the structural-validity cases and the two soundness denominators, which is the correct split |
| Cycle detection removed | exactly the 3 cycle tests red |
| Kernel delegation replaced with blind trust | exactly the 2 tests that depend on it red |

`--project process` fails on this machine for an environment reason — it requires the
built binary linked on `PATH` — and is unrelated to this change.


rsalus:
@coderabbitai approve

PR #1906, opened 2026-09-12
feat(settle): adjudicate one batch against the capsule that was pinned when the work was compiled

rsalus:
## Summary

The second half of the semantic plane's endpoint, stacked on #1905. `prepare`
compiles a capsule, the harness executes it with zero governance calls, and
`settle` adjudicates the claims that come back **against the capsule the work ran
under** — never against whatever the design says now. That clause is the whole
point of pinning, and it is why this action takes the compiled artifact as an
argument rather than taking a workflow id and going looking for terms.

> Stacked on #1905. Review that one first; this branch merges it.

### What it decides

Each claim is read against the capsule's declared result shape for its task,
evidence against its admitted kinds, deviations against its envelope, and the
batch against `settlementContract.requiredResults`. Ten finding kinds, every one
exercised by a case, with a roster the tests iterate so a kind added without a
case reddens.

Nine of the ten refuse the batch. `deviation-awaiting-approval` **holds** it
instead, and that asymmetry is deliberate: a deviation inside the envelope is a
legitimate answer from a worker that found the capsule's assumptions wrong, and
refusing it would push that worker toward silently complying with a premise it
has already disproved.

**A rejected batch is a successful call.** It returns the findings and appends the
record, because the findings are the caller's next input and the next call has to
be able to read that this batch did not take. Only three things answer with an
error, and none of them adjudicates anything: a malformed request, a capsule that
fails the published contract (`CAPSULE_INVALID`), and a capsule that parses but
whose own references do not resolve (`CAPSULE_UNRESOLVED`). The last two are
separated because they are completely different to fix — and because a capsule
that parses and does not resolve is exactly the document a structural check alone
waves through.

### The executor question, answered on merits

The charter asked for a clean break over tech debt. `settle` runs nothing, so the
executor's compiler and run loop had nothing to offer it — it was written
greenfield. What the two share is the substrate underneath, and that is imported
rather than reimplemented: content-addressed custody, canonical encoding, the
operation claim, the settlement-endpoint oracle. `execute_intent` is untouched.

Two helpers moved rather than being copied a second time. `outerCorrelation` and
`stampFromAmbient` were private to the executor and both verbs need them; a second
hand-written copy is how the correlation defect the first one fixed comes back for
whichever verb gets the copy slightly wrong.

Custody order is the executor's, for the executor's reason: the adjudication
interior reaches content-addressed storage **before** the record naming it commits,
so a failed bundle write fails the whole settlement rather than leaving a row the
integrity oracle condemns on sight.

### Three surfaces where measurement beat the plan

- **The emitter-closure census could not see the append.** A literal `type:` was
  tried first to keep it visible. It does not work — that census resolves
  `.append(...)` call sites and this commit goes through `decideOnce` — so the
  literal bought nothing and cost the endpoint constant's guarantee. The allowance
  row goes in, and the list's rationale, which described only the
  runtime-discriminant case, now names both mechanisms.
- **`settle` is probed, not excluded.** Everything excluded from the shipped-emitter
  corpus shells out, touches a worktree, or resolves a host repository; this action
  does none of those. The probe runs the *accepting* path against the capsule
  contract's own fixture, so a schema change reaches it rather than leaving it
  asserting against a shape that no longer ships.
- **The packaged sweep reports 126/126 actions reachable through the compiled
  binary**, so the action is exercised end to end and not only in process.

## Changes

- **`src/verbs/settle/`** — `adjudicate.ts` (pure: no store, no clock),
  `settlement-bundle.ts` (the custody document), `handler.ts`, plus types, output
  schema and a measured economy budget.
- **`src/registry/actions/orchestrate/settle.ts`** — the registration. `outputSchema`
  is `withCappedShape(...)`, which is the only route open to a new action: the
  vacuity waiver ids are a compile-enforced shrink-only union.
- **`execution.settled`** across every surface — catalog, data schema, inferred type,
  `EVENT_DATA_SCHEMAS`, `EventDataMap`, annotation, the canonical projection case
  and its `never` trap, and the `SETTLEMENT_ENDPOINTS` row, which is added only now
  because an endpoint listed ahead of its producer would condemn the first record
  that producer ever writes.
- **`src/dispatch/core/outer-correlation.ts`** — the two shared correlation helpers.
- **Authority lock re-approved.** The action set moved, so the `action-id-registry`
  digest moved. The note also corrects its own previous sentence claiming nothing
  imports `strategos-contracts` — the capsule contract does, and `settle` is that
  contract's first production consumer, which is why both capsule modules lose their
  `RESERVED` headers here.

Riders, each a citation that resolves to nothing until it is fixed:

- `verbs/execute/types.ts` and `run-bundle.ts` both said the workflow kernel "does
  not exist yet". It shipped. The reason lowering is still deferred has changed from
  "no target" to "nothing produces the input", and that is what they say now.
- `contract/ir/admission-ir-schema.ts` called its drift guard "co-located" when every
  test in this repository lives under `tests/`.
- `contract/reachability/README.md` and the packaged-proof baseline note both named
  regeneration commands inside `servers/exarchos-mcp`, a package that was dissolved.
- The mutation-adequacy roster's title said 83 while asserting 84.

Pins and artifacts: 125 → 126 actions, 83 → 84 on `exarchos_orchestrate`, 180 → 181
event types, primary-owner population 71 → 72. Generated artifacts regenerated by
their own commands. Inventory rows spliced, not regenerated — and five of the ten
spliced rows were **stale on main** rather than moved by this change.

## Test Plan

- `npm run typecheck` — clean across all three projects
- `npm run lint` — exit 0
- `npx vitest run --project core` — **13,313 passed**, 0 failed
- `npx vitest run --project unit` — **1,165 passed**, 0 failed
- `npm run test:conformance` — **328 passed**
- `npx vitest run --project acceptance` — **108 passed**
- `npx vitest run --project outcome` — **32 passed**
- Every CI-only gate derived from `ci.yml` (36 commands) run locally — all pass
- The wire golden regenerated deliberately, the diff measured **per tool and per
  field** rather than by line — one tool changed, three additive surfaces, no
  existing action's digest moved — and recorded in that file's header as its
  convention requires

Kill probes on the new modules, each biting precisely:

| Probe | Result |
|---|---|
| Adjudication always settles | 17 of 54 red |
| Reference resolution skipped in the handler | exactly the `CAPSULE_UNRESOLVED` case |
| Replay pre-flight removed | **initially green — the test was too weak** |

That third probe is worth reading. The replay test compared receipts, and
`decideOnce` returns the first call's result either way, so a replay that fell
through to the effect returned an identical receipt. What it also did was build a
second bundle — a fresh `settledAt` makes a different digest — and leave it in
custody as an orphan. The test now counts blobs in the bundle store, and the probe
bites.


rsalus:
@coderabbitai approve

